PR #19, opened 2021-03-27
Changed the startup animation

ef1500:
Load

Was previously a pull request [here](https://github.com/manga-download/hakuneko/pull/3231).

### DIFF
--- a/src/css/cubeloader.css
+++ b/src/css/cubeloader.css
@@ -1,0 +1,267 @@
+body {
+    font-family: monospace;
+    font-size: 150px;
+    background-color: #202225;
+}
+
+.container {
+    display: flex;
+    justify-content: center;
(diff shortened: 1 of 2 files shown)


Sheepux:
Feature lost: 
- Button to close the loader, because it's currently not doing anything 😄 

Also i'm not fond of externalising CSS when there is no plan to reuse it but that might be a personal taste.

ronny1982:
I think it is a good idea to embed the CSS styles into the HTML file, so there is just one request to show the splash instead of two

Sheepux (changes requested):
two small improvements identified

Sheepux:
Merged with a custom implementation.
Changed a bit the colors to the red one.
Final result below. It's nice !
hakuneko-loading2

Issue #3, opened 2020-11-20
mockup plugin selector


Sheepux:
- [x] Dropdown
- [x] Advanced selection/filter
idea by @devemloop
![image](https://user-images.githubusercontent.com/19878135/99855453-33870c80-2b87-11eb-8153-35e1e03e47be.png)

yoshiyes:
![image](https://user-images.githubusercontent.com/20050570/116631870-1d5f5280-a956-11eb-80dc-12cb49948672.png)
WIP

PR #21, opened 2021-05-09
Plugin selector UI update

yoshiyes:
Refer to #3


src/frontend/sample-svelte/components/MediaSelect.svelte
```diff
@@ -67,7 +67,7 @@
         });
     }
 
-    let showPluginSelect = false;
+    let myPluginModalOpen = false;
```
Sheepux, line 70: prefer isPluginModalOpen

Sheepux (changes requested):

Sheepux (approved):

PR #27, opened 2021-06-30
feat(front): add smartImageLoader svelte component

Pierre-Monier:
I remove the unrelated changes


Sheepux:
Merging, but i'll rework it so that all the components are typescripted <script lang="ts">

Issue #15, opened 2020-12-11
mockup tutorial


Sheepux:
Implement the "how to use hakuneko" pages accessible from the home screen

Sheepux:
Last commit introduced a router that will allow to write some guides.

Pierre-Monier:
Can we actually make a raw wireframe for content that should appear on each "tutorial item" ? Also the disposition is a bit weird I think ? Is that desired ?

![Screenshot 2023-11-24 at 14 15 54](https://github.com/manga-download/haruneko/assets/65488471/a5019e02-9c4b-4818-a7bf-8c5434a7d952)

Sheepux:
I'm currently working on it. I plan to generate a remote page as in current documentation: https://hakuneko.download/docs/interface/
The HTML would be fetched and split into each page and integrated with carbon's css.
Yeah the strange split was intentional at first (i was thinking in "sections" but wanted to work on this later)

Sheepux:
documentation generator is here:: https://github.com/manga-download/manga-download.github.io/tree/master/_docs

Pierre-Monier:
Oh, ok I see, seems like a good idea :). Maybe we can add carbon css to the documentation website and just use iframe for the app ? Might be cool if documentation as the same UI as the app :)

Sheepux:
PR https://github.com/manga-download/haruneko/pull/454 introduced remote documentation fetching from  https://hakuneko.download/docs/haruneko/ (markdown of the webpage  https://github.com/manga-download/manga-download.github.io/blob/master/_docs/06-HaruNeko.md )
What's left is putting actual content in the documentation (and extra todos discussed in the PR)

PR #28, opened 2021-07-03
fix(front/home): add better home page style

Pierre-Monier:
I think it close this [issue](https://github.com/manga-download/haruneko/issues/15)

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "npm": ">=7.6.0"
   },
   "dependencies": {
-    "fs-extra": "9.0.1"
+    "fs-extra": "9.0.1",
+    "svelte-awesome": "^2.3.2"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "latest",
--- a/package.json
+++ b/package.json
@@ -10,8 +10,7 @@
     "npm": ">=7.6.0"
   },
   "dependencies": {
-    "fs-extra": "9.0.1",
-    "svelte-awesome": "^2.3.2"
+    "fs-extra": "9.0.1"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "latest",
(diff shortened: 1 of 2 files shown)


Sheepux:
The home page is not the tutorial i have in mind (but it's the starting point).
I'll have to write the use case in the issues as the titles were only a way to dump what was on my mind.
/LGTM

package.json
```diff
@@ -10,7 +10,8 @@
     "npm": ">=7.6.0"
   },
   "dependencies": {
-    "fs-extra": "9.0.1"
+    "fs-extra": "9.0.1",
+    "svelte-awesome": "^2.3.2"
```
Sheepux, line 14: Do not use svelte awesome,  use svelte carbon icons so that the icons stays coherents

Sheepux: The font awesome code was just a copy/paste legacy from hakuneko :)

Pierre-Monier: Ok got it :)

src/frontend/sample-svelte/components/Home.svelte
```diff
@@ -1,46 +1,93 @@
...
-        border-radius: 10%;
-    }
-</style>
+<script>
+    import Search16 from "carbon-icons-svelte/lib/Search16";
+    import Plug16 from "carbon-icons-svelte/lib/Plug16";
+    import Language16 from "carbon-icons-svelte/lib/Language16";
+    import Renew16 from "carbon-icons-svelte/lib/Renew16";
+
+    const iconsStyle = "margin-left: 5px; margin-right: 5px;";
```
Sheepux, line 7: I think that using class is more consistent with styling than using jsstyles:

```html
<style>
  :global(svg.custom-class) {
    fill: blue;
  }
</style>

<Add16 class="custom-class" />
```

Sheepux (approved):

Issue #12, opened 2020-12-09
Viewer throttling


Sheepux:
Something like
https://codepen.io/ycpand/pen/mwrZXg

Pierre-Monier:
I can't see the link, I got a 403 error in the console. I have a non-premium account on CodePen, not sure if it's related

Sheepux:
I don't have any account, it's open to everyone.

here is the codepen:
`<div class="container">
  <img src="http://cumbrianrun.co.uk/wp-content/uploads/2014/02/default-placeholder-300x300.png" alt="" data-src="https://instagram.ftpe8-4.fna.fbcdn.net/vp/858878e86ee732def268b2b33fadb5bc/5B7554EE/t51.2885-15/e35/30078878_134250340653099_1383951987346767872_n.jpg" />
  
  <img src="http://cumbrianrun.co.uk/wp-content/uploads/2014/02/default-placeholder-300x300.png" alt="" data-src="https://instagram.ftpe8-4.fna.fbcdn.net/vp/858878e86ee732def268b2b33fadb5bc/5B7554EE/t51.2885-15/e35/30078878_134250340653099_1383951987346767872_n.jpg" />
  
  <img src="http://cumbrianrun.co.uk/wp-content/uploads/2014/02/default-placeholder-300x300.png" alt="" data-src="https://instagram.ftpe8-4.fna.fbcdn.net/vp/858878e86ee732def268b2b33fadb5bc/5B7554EE/t51.2885-15/e35/30078878_134250340653099_1383951987346767872_n.jpg" />
  
  <img src="http://cumbrianrun.co.uk/wp-content/uploads/2014/02/default-placeholder-300x300.png" alt="" data-src="https://instagram.ftpe8-4.fna.fbcdn.net/vp/858878e86ee732def268b2b33fadb5bc/5B7554EE/t51.2885-15/e35/30078878_134250340653099_1383951987346767872_n.jpg" />
  
  <img src="http://cumbrianrun.co.uk/wp-content/uploads/2014/02/default-placeholder-300x300.png" alt="" data-src="https://instagram.ftpe8-4.fna.fbcdn.net/vp/858878e86ee732def268b2b33fadb5bc/5B7554EE/t51.2885-15/e35/30078878_134250340653099_1383951987346767872_n.jpg" />
  
[...]

PR #30, opened 2021-07-08
feat(front/viewer): add throttling first implementation

Pierre-Monier:
Trying to solve [this issue](https://github.com/manga-download/haruneko/issues/12), need feedback


src/frontend/classic/components/Viewer.svelte
```diff
@@ -40,11 +44,81 @@
...
+    function loadWide(el: Element) {
+        const img = el as HTMLImageElement;
+        const src = img.getAttribute("data-image");
+
+        if (src) {
+            img.src = src;
+        }
+    }
+
+    function loadImages(currentIndex: number): void {
```
Sheepux, line 70: Should use a simple foreach with await instead of a recursive function for a sequential processing.

src/frontend/classic/components/Viewer.svelte
```diff
@@ -40,11 +44,81 @@
...
+
+            if (currentIndex + 1 < imagesToLoad.length) {
+                loadImages(currentIndex + 1);
+            } else {
+                imagesToLoad = [];
             }
+        }, throttlingDelay);
+    }
+
+    function isImagesToLoadFull(): boolean {
```
Sheepux, line 88: there is no need to create such small function if there is no plan to reuse it.

src/frontend/classic/components/Viewer.svelte
```diff
@@ -40,11 +44,81 @@
...
+            const index = Number(el.getAttribute("data-index"));
+            imagesToLoad[index] = el;
         }
+
+        if (isImagesToLoadFull()) {
+            loadImages(0);
+        }
+    }
+
+    function isThumbnailImageUtil(el: HTMLElement): boolean {
```
Sheepux, line 103: there is no need to create such small function if there is no plan to reuse it.

src/frontend/classic/components/Viewer.svelte
```diff
@@ -96,20 +170,33 @@
                         <p>...loading image</p>
                     {:then content}-->
                     {#if mode === "Thumbnail"}
+                        <img
+                            src={placeholderImage}
+                            alt=""
+                            class="thumbnail-image-util"
+                            use:processImage={index}
```
Sheepux, line 177: why not use on:load instead of use if there is no additional behavior extended  ?

src/frontend/classic/components/Viewer.svelte
```diff
@@ -40,11 +44,81 @@
...
+        }, throttlingDelay);
+    }
+
+    function isImagesToLoadFull(): boolean {
+        return imagesToLoad.length === item.Entries.length;
+    }
+
+    function addElement(el: Element | null) {
+        if (el && el.getAttribute("data-index")) {
+            const index = Number(el.getAttribute("data-index"));
```
Sheepux, line 94: i see a lot of Dom processing but it could be handled without an index by externalising each image/thumbnail into components that can handle their own context.
I think it would simplify the code

Sheepux (changes requested):
Look ok for a starting point
A few things to change before merging
I'll probably try to change the way the code it split.

Regarding the preload i've found something that seems more "svelte like"
https://svelte.dev/repl/64b6e15234594119a73ec45513674fca?version=3.38.3

src/frontend/classic/components/Viewer.svelte
Pierre-Monier: I just think it's more readable this way

src/frontend/classic/components/Viewer.svelte
Pierre-Monier: Same here, the clean code book recommend using this kind of small function but I'm ok with removing it

src/frontend/classic/components/Viewer.svelte
Pierre-Monier: It's absolutely right

src/frontend/classic/components/Viewer.svelte
```diff
@@ -6,22 +6,26 @@
...
     import { createEventDispatcher } from "svelte";
-    const dispatch = createEventDispatcher();
-
     import type { IMediaContainer } from "../../../engine/providers/MediaPlugin";
 
+    type Mode = "Thumbnail" | "Wide" | "Video";
+
     export let item: IMediaContainer;
+    export let mode: Mode = "Thumbnail";
+    export let throttlingDelay: number = 1000;
```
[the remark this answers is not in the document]
Pierre-Monier: Yes, I wasn't aware of that at that time

Pierre-Monier:
I will work on that soon

src/frontend/classic/components/Viewer.svelte
Sheepux: Yes and No, for example, i had to scroll up to find the test done and go back. That was harder to read ;)
Alternative is to name the test within a variable
`const isThumbnailImageUtil: Boolean = myval-1 > 0;`

Pierre-Monier: Ok I see your point ^^, thanks for the teaching :)

Sheepux:
I just remembered that i saw a reader project in JS that we could try to integrate as the viewer: look at #31

Pierre-Monier:
I improve it with review feedback, I think it can be merged now. I will take a deeper look at OpenReader to see what we can do with it :)

Sheepux:
make sure you mark conversations as resolve once you've solved them ;)

Sheepux (approved):
Sorry for the delay.
Merged. 
I'll probably fuse the "ViewerThumbnailImage" into a  single component with a "mode" prop instead of having 2 components.

PR #32, opened 2021-08-13
Feat/app menu

Pierre-Monier:
First app menu implementation. The UI tab is working though. The menu is construct as follows : 

```
-General
--General settings
-UI
--UI settings
-Websites
--Website1
---Website1 settings
--Website2
---...
-Help
--Help links
-About
--About data
```

I think it's a good simple structure, close to what we currently got on hakuneko
I think it's a good start and with this we can start to implement more logic (persist setting, link setting to request...).
Maybe later if it's needed we can implement some modal in the UI.


Pierre-Monier:
Settings value are now persisted in localStorage

src/frontend/classic/utils/storage.ts
```diff
@@ -0,0 +1,46 @@
+export type SettingValue = string | number | boolean;
+export type SettingExtras = ((extrasArg?: SettingValue) => void) | undefined
+export type SettingValidator = ((value: SettingValue) => boolean) | undefined
+
+export const onSettingChange = (
```
Sheepux, line 5: Use svelte stores. Probably one dedicated to each kind ? ("General","UI").

```typescript
import { writable } from "svelte/store";

const storedValue= localStorage.getItem("thisvalue");
export const readthisvalue = writable(storedValue);
readthisvalue .subscribe(value => {
    localStorage.setItem("thisvalue", value === 'mooo' ? 'mooo' : 'meh');
});
```

src/frontend/classic/App.svelte
```diff
@@ -66,9 +44,13 @@
     // UI
     let isSideNavOpen = false;
     let isOpen = false;
-    let theme: string;
+    let theme = getSettingDefaultValue(storageKeys.THEME, "hakuneko") as string;
```
Sheepux, line 47: Just set theme to a svelte/store, it'll be easy to store/modify/subscribe and no need to pass it as prop (see the svelte/store comment further)

src/frontend/classic/components/MenuLeftPanel.svelte
```diff
@@ -0,0 +1,283 @@
...
+    export let changeUIMode: () => void;
+    export let changeTheme: (themeId: string) => void;
+
+    function openExternalLink(uri: string) {
+        nw.Shell.openExternal(uri);
+    }
+
+    const win = nw.Window.get();
+
+    $: sideNavStyle = isSideNavOpen ? `min-width: ${win.width / 2}px;` : "";
```
Sheepux, line 34: Feels too wide. Wasn't the default value enougth ? I feel like the global UI flow is now disturbed.

src/frontend/classic/App.svelte
```diff
@@ -96,70 +82,14 @@
...
-            />
-            <HeaderGlobalAction
-                on:click={() => win.close()}
-                aria-label="Close"
-                icon={CloseIcon}
-            />
-            <HeaderAction bind:isOpen>
-                <HeaderPanelLinks>
-                    <HeaderPanelDivider>Interface</HeaderPanelDivider>
-                    <HeaderPanelLink>
```
Sheepux, line 130: I'm sad that the UI settings aren't on the right side anymore, it was nice to test the UI changes. The menu is hiding everything. Could be in both places.

Sheepux (commented):
Early review. I'll test more next week.
I think you have to dig into svelte/stores :)

Sheepux:
Contributes to #8

package.json
```diff
@@ -10,7 +10,8 @@
     "npm": ">=7.6.0"
   },
   "dependencies": {
-    "fs-extra": "latest"
+    "fs-extra": "latest",
+    "uuid": "^8.3.2"
```
ronny1982, line 14: Avoid additional dependencies, a weak UUID can probably be created with a single line of JavaScript.
I don't see a reason for a cryptographically strong UUID.

scripts/build-app.mjs
```diff
@@ -12,7 +12,9 @@ async function createApplicationManifest() {
     const manifest = {
         name: pkgConfig.title,
         description: pkgConfig.description,
-        'chromium-args': '--ignore-certificate-errors',
+        // --enable-automation is use to disable password saving bubble
+        // https://github.com/GoogleChrome/chrome-launcher/issues/32
+        'chromium-args': '--ignore-certificate-errors --enable-automation',
```
ronny1982, line 17: Why is it necessary to disable password popup?
I was eventually planning to use this so users can store their account data and login to websites automatically.

src/frontend/classic/App.svelte
```diff
@@ -96,70 +82,14 @@
...
+    <AppBar
+        {isSideNavOpen}
+        {isOpen}
+        {winMaximized}
+        {changeUIMode}
+        {changeTheme}
+    />
 
     <Content id="hakunekoapp">
         <MediaSelect on:select={(e) => (selectedMedia = e.detail)} />
```
ronny1982, line 95: May remove redundant brackets `e => ...` instead of `(e) => ...`.
May use more expressive variable names, e.g. `evt` or `event` instead of `e`

src/frontend/classic/theme/sidenav-hack.css
```diff
@@ -0,0 +1,58 @@
+.bx--side-nav {
```
ronny1982, line 1: Why is this file called hack?
Seems to me like overwriting pre-defined classes from a theme, so i suggest a more appropriate name.

src/frontend/classic/utils/storage.ts
```diff
@@ -0,0 +1,46 @@
...
+    key: string,
+    value: SettingValue,
+    validator: SettingValidator,
+    extras: SettingExtras,
+    extrasArg: SettingValue | undefined
+): boolean => {
+    if (!validator || (validator && validator(value))) {
+        localStorage.setItem(key, value.toString());
```
ronny1982, line 13: The `localStorage` is assumed not to be persistent and may (theoretically) be cleared whenever chrome feels like it. HakuNeko engine will provide certain levels of storages in the future...

src/frontend/classic/theme/sidenav-hack.css
Sheepux: It's a valid temporary fix because carbon (the base ibm project) doesn't allow theme customisation on the sidenav. It will live untill they code it.

src/frontend/classic/utils/storage.ts
Sheepux: I've review and it should be shifted to svelte/store that will be linked to hakuneko storage (but will be linked to local storage while we're moking).

src/frontend/classic/components/FolderSelectorInput.svelte
```diff
@@ -0,0 +1,44 @@
...
+    import { v4 as uuidv4 } from "uuid";
+
+    export let labelText = "Select directory";
+    export let storageKey: string | undefined = undefined;
+    export let extras: SettingExtras = undefined;
+    export let validator: SettingValidator = undefined;
+    export let passNewValueToExtras: boolean = false;
+    export let path: string = "";
+
+    const uuid = uuidv4();
```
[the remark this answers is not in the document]
Pierre-Monier: Yes we can pass a selfdefiningID as a prop. The idea was to abstract this logic inside the component, so we don't have to handle this when calling the FolderSelectorInput component. But yeah, it might be overkill, and adds an extra dependency. I will fix this

src/frontend/classic/utils/storage.ts
Pierre-Monier: Ok, I wasn't sure if we wanted to use svelte store, so I pick up the easiest option ^^

src/frontend/classic/components/MenuLeftPanel.svelte
Pierre-Monier: It's too large when the window is extended, but when you have a small window it wasn't wide enough

src/frontend/classic/App.svelte
Pierre-Monier: I like to have one thing in one place I think it's easier but yes we can set it in two places for testing

package.json
Pierre-Monier: I agree that it's a little overkill

scripts/build-app.mjs
Pierre-Monier: Oh ok, sounds like a neat feature. I was just thinking that having this pop up will be annoying but ok

src/frontend/classic/theme/sidenav-hack.css
Pierre-Monier: Like @Sheepux said, it's a temporary fix so for me the name is appropriate

src/frontend/classic/App.svelte
Pierre-Monier: I change `e` to `evt`, but the brackets can't be removed with the current linter config

src/frontend/classic/components/MenuLeftPanel.svelte
Pierre-Monier: Ok my bad I remove the style and it's working nice

Pierre-Monier:
I integrate the svelte/store feature. I decide to split every single setting into his own store. With this component will be updated only when the setting used is updated.

Sheepux:
I guess we can go merging and commiting while we're still moking the interface.
I'm not much in the mood to review currently (work overload). So if needed we'll rework later.

PR #51, opened 2022-07-31
itemmarkManager

Sheepux:


src/engine/HakuNeko.ts
```diff
@@ -48,6 +51,9 @@ export class HakuNeko {
         return this.#bookmarkPlugin;
     }
 
+    public get ItemmarkManager(): ItemmarkManager {
```
ronny1982, line 54: Maybe rename to `ItemcheckManager` or `ItemflagManager`

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
+import type { IMediaContainer } from './providers/MediaPlugin';
+import { type StorageController, Store } from './StorageController';
+
+export class ItemmarkManager {
+
+    private readonly Items:Map<string,ItemMark[]>=new Map();
```
ronny1982, line 6: Use spaces to improve readability

ronny1982: Non-public fields may be camel-case

src/frontend/classic/components/MediaItemSelect.svelte
```diff
@@ -89,7 +92,29 @@
...
+                    selectedItems[0],
+                    MarkType.Viewed
+                );
+                $selectedItem = selectedItems[0];
+            }}
+        />
+        <ContextMenuOption
+            indented
+            labelText="MarkAsViewed"
+            shortcutText="⌘V"
```
ronny1982, line 110: Shortcut already used in other context menu option

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+        const marks = this.Find(entry.Parent) ?? [];
+        marks.push(newmark);
+        const key = this.StorageKey(entry.Parent);
+        this.Items.set(key,marks);
+        await this.storage.SavePersistent<ItemmarkSerialized>(this.Serialize(entry, marks), Store.Itemmarks, key);
+        return marks;
+    }
+
+    /* Remove the mark of an item */
+    public async Remove(entryToRemove: IMediaContainer): Promise<ItemMark[]> {
```
ronny1982, line 49: Caller probably has the mark at hand that shall be removed, return a list is not expected
`Remove(mark: ItemMark): Promise<bool>`

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+        await this.storage.SavePersistent<ItemmarkSerialized>(this.Serialize(entryToRemove, marks), Store.Itemmarks, key);
+        return marks;
+    }
+
+    /* Key to store marks by plugin-media */
+    private StorageKey(item: IMediaContainer): string {
+        return `${item.Parent.Identifier} :: ${item.Identifier}`;
+    }
+
+    public Hash(text: string): string {
```
ronny1982, line 69: Should be private

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+        });
+        const key = this.StorageKey(entryToRemove.Parent);
+        this.Items.set(key,marks);
+        await this.storage.SavePersistent<ItemmarkSerialized>(this.Serialize(entryToRemove, marks), Store.Itemmarks, key);
+        return marks;
+    }
+
+    /* Key to store marks by plugin-media */
+    private StorageKey(item: IMediaContainer): string {
+        return `${item.Parent.Identifier} :: ${item.Identifier}`;
```
ronny1982, line 66: Parent seems redundant as a list of markers is already stored related to provider and container ID (see line 14)
The `item.Identifier` may be long, may use `Hash(item.Identifier)` to save storage

src/engine/ItemmarkManager.ts
Sheepux: this allows the UI that is using the list to not refilter the marks by simply reassigning the results already filtered

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public Hash(text: string): string {
+        return text.split('').reduce((hash, c) => 31 * hash + c.charCodeAt(0) | 0, 0).toString(36);
+    }
+
+    /* Get the mark of an item */
+    public getMark(item: IMediaContainer, marks: ItemMark[]): MarkType{
```
ronny1982, line 74: Parameter `marks` should not be required, the `ItemmarkManager` must be able to know its corresponding `marks`

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+            Media: {
+                ProviderID: media.Parent.Parent.Identifier,
+                EntryID: media.Parent.Identifier
+            },
+            Marks: marks,
+        };
+    }
+
+    /* Retrieves the marks of a media */
+    public Find(entry: IMediaContainer): ItemMark[] {
```
ronny1982, line 29: Is this public by intention? Seems more like it was meant to be private

src/engine/ItemmarkManager.ts
Sheepux: I forgot to re-add it as optional (as the first line suggests).
I wanted to reduce the quantity of call on the Item.get as the list could be growing (don't know the performance of Map.get).
The list has already been retrieved by the UI and can be used to gain a step.
This was my thiking about performances

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+    private async Load() {
+        const mediamarks = await this.storage.LoadPersistent<ItemmarkSerialized[]>(Store.Itemmarks);
+        mediamarks.forEach(itemmark => {
+            this.Items.set(`${itemmark.Media.ProviderID} :: ${itemmark.Media.EntryID}`,itemmark.Marks);
+        });
+    }
+
+    private Serialize(media: IMediaContainer, marks: ItemMark[]): ItemmarkSerialized {
+        return {
+            Media: {
```
ronny1982, line 20: May be simplified to
```javascript
MediaID: `${itemmark.Media.ProviderID} :: ${itemmark.Media.EntryID}`
```

src/engine/ItemmarkManager.ts
Sheepux: Yes, the UI fetches the marks of a media, and caches it as the UI will call the getMarks a lot
Used here: https://github.com/manga-download/haruneko/pull/51/files#diff-cc54b4e21d800248a19197cf1c4d48632d290d108ec4f403a3fa137e89679ac7R72

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export class ItemmarkManager {
+
+    private readonly Items:Map<string,ItemMark[]>=new Map();
+    constructor(private readonly storage: StorageController) {
+        this.Load();
+    }
+
+    private async Load() {
+        const mediamarks = await this.storage.LoadPersistent<ItemmarkSerialized[]>(Store.Itemmarks);
```
ronny1982, line 12: Can't this be loaded directly without (de-)serialization? `this.Items = await this.storage.LoadPersistent<ItemmarkSerialized[]>(Store.Itemmarks);`

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+        const newmark: ItemMark = {
+            IdentifierHash: this.Hash(entry.Identifier),
+            TitleHash: this.Hash(entry.Title),
+            kind: kind,
+        };
+        const marks = this.Find(entry.Parent) ?? [];
+        marks.push(newmark);
+        const key = this.StorageKey(entry.Parent);
+        this.Items.set(key,marks);
+        await this.storage.SavePersistent<ItemmarkSerialized>(this.Serialize(entry, marks), Store.Itemmarks, key);
```
ronny1982, line 44: Can't this be stored directly without serialization?
`await this.storage.SavePersistent<ItemmarkSerialized>(this.Items, Store.Itemmarks, key);`

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+        if (!marks) return;
+        marks = marks.filter(entry => {
+            const doesIdentifierMatch = entry.IdentifierHash === this.Hash(entryToRemove.Identifier);
+            const doesTitleMatch = entry.TitleHash === this.Hash(entryToRemove.Title);
+            return !doesIdentifierMatch && !doesTitleMatch;
+
+        });
+        const key = this.StorageKey(entryToRemove.Parent);
+        this.Items.set(key,marks);
+        await this.storage.SavePersistent<ItemmarkSerialized>(this.Serialize(entryToRemove, marks), Store.Itemmarks, key);
```
ronny1982, line 60: Duplicated code (see `Add`), consider extract to shared private method

src/engine/ItemmarkManager.ts
Sheepux: If we save as map directly, it would be Loaded back as a map ?

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /* Retrieves the marks of a media */
+    public Find(entry: IMediaContainer): ItemMark[] {
+        return this.Items.get(this.StorageKey(entry));
+    }
+
+    /* Add a mark of an item */
+    public async Add(entry: IMediaContainer, kind: MarkType): Promise<ItemMark[]> {
+        const newmark: ItemMark = {
```
ronny1982, line 35: Should not be allowed to add `MarkType.None` => save storgae space

Sheepux: None isn't stored, it's returned by getMark if none ... but i could have returned null instead ... #Brain

src/engine/ItemmarkManager.ts
ronny1982: Consider better method name as `GetMark()` implies to return a `ItemMark`

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+            IdentifierHash: this.Hash(entry.Identifier),
+            TitleHash: this.Hash(entry.Title),
+            kind: kind,
+        };
+        const marks = this.Find(entry.Parent) ?? [];
+        marks.push(newmark);
```
ronny1982, line 41: Must not add a mark that already exist (ID, Title)

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+        return text.split('').reduce((hash, c) => 31 * hash + c.charCodeAt(0) | 0, 0).toString(36);
+    }
+
+    /* Get the mark of an item */
+    public getMark(item: IMediaContainer, marks: ItemMark[]): MarkType{
+        marks = marks || this.Find(item.Parent);
+        if (!marks) return MarkType.None;
+        const itemIdentifierHash = this.Hash(item.Identifier);
+        const itemTitleHash = this.Hash(item.Title);
+        const mark = marks.find(mark => {
```
ronny1982, line 79: Duplicated code (see `Remove()`), consider extract shared code to private method

src/engine/ItemmarkManager.ts
ronny1982: Consider using a more expressive method name

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+            const doesIdentifierMatch = mark.IdentifierHash === itemIdentifierHash;
+            const doesTitleMatch = mark.TitleHash === itemTitleHash;
+            return doesIdentifierMatch || doesTitleMatch;
+        });
+        return mark ? mark.kind : MarkType.None;
+    }
+}
+
+type ItemmarkSerialized = {
```
ronny1982, line 88: No longer required in case of ditching serialization

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+    Media: Provider,
+    Marks : ItemMark[],
+}
+type Provider = {
```
ronny1982, line 92: No longer required in case of ditching serialization

src/engine/ItemmarkManager.ts
ronny1982: Also consider an equals operator/method to determine if two marks are representing the same media item

src/frontend/classic/components/MediaItemSelect.svelte
```diff
@@ -8,26 +8,28 @@
...
     import { selectedMedia, selectedItem } from '../Stores';
 
-    import type { IMediaContainer } from "../../../engine/providers/MediaPlugin";
-    
+    import type { IMediaContainer } from '../../../engine/providers/MediaPlugin';
+    import { ItemMark, MarkType } from '../../../engine/ItemmarkManager';
 
     let items: IMediaContainer[] = [];
     let filteredItems: IMediaContainer[] = [];
+    let marks: ItemMark[] = [];
```
ronny1982, line 24: Variable not really nedded

Sheepux: marks is used as a cache to reduce the amount of Items.get on the manager ... but is that reallly needed ?
`mark={HakuNeko.ItemmarkManager.getMark(item, marks)}`
see comment : https://github.com/manga-download/haruneko/pull/51#discussion_r934261511

ronny1982: If this is intended to be used as observable, consider finding a different approach to update UI

src/engine/ItemmarkManager.ts
```diff
@@ -0,0 +1,105 @@
...
+        };
+    }
+
+    /* Retrieves the marks of a media */
+    public Find(entry: IMediaContainer): ItemMark[] {
+        return this.Items.get(this.StorageKey(entry));
+    }
+
+    /* Add a mark of an item */
+    public async Add(entry: IMediaContainer, kind: MarkType): Promise<ItemMark[]> {
```
ronny1982, line 34: Return type is not expected, should be void or bool.
If the caller is interested in the updated list, there should be a dedicated method (e.g. `Find()` but with more expressive name)

ronny1982:
Make sure to merge `main` before pushing any other changes to this branch, so the website tests (1.5 hours) are not executed during a pull request.

Consider adding [unit](https://github.com/manga-download/haruneko/blob/master/src/engine/PluginController_test.ts) and/or [integration](https://github.com/manga-download/haruneko/blob/master/src/engine/SettingsGlobal_e2e.ts) tests for `ItemmarkManager`, it is likely there are some edge test cases not yet discovered in the review...

src/engine/ItemmarkManager.ts
ronny1982: If performance is a concern, avoid caliing the `getMark` method at all from UI (it calculates hashes for each call).
Consider providing a map instead and making `getMark` private

e.g.
```
// Create a map that holds the mark for each chapter of the given manga
Find(container) {
    const cache = new Map<IMediaContainer, MarkType>();
    const marks = this.Items.get(this.StorageKey(entry));
    for(const entry of container.Entries) {
        cache.set(entry, this.getMark(entry));
    }
    return cache;
}
```

Sheepux: one of the main "issue" with svelte is that it uses assign statement (=) to update a component, but i'll try to find if i can get an alternative way to forces it to detect a change on a map element (I read something about keyed elements in svelte)

src/engine/ItemmarkManager.ts
ronny1982: I see the intention, but business logic shoud be genral purpose, just ignore the fact that it will b consumed by a frontend.
May introduce antoher method to acquire the updated list of marks.

src/engine/ItemmarkManager.ts
ronny1982: If it doesn't work use a `Record<string, MarkItem[]>` instead

ronny1982: > If we save as map directly, it would be Loaded back as a map ?

Good catch, this may solv this if keeping `Map`
```javascript
const map = new Map([ [ 'a', 1 ], [ 'b', 2 ] ]);
const serialized = [ ...map.entries() ];
const deserialized = new Map(serialized);
```

Sheepux:
> Make sure to merge `main` before pushing any other changes to this branch, so the website tests (1.5 hours) are not executed during a pull request.

Did you mean to rebase from master or to commit directly into master so that it does not trigger any github action ?

PR #64, opened 2023-01-17
Common: Fix search params & Add MangasMultiPageCSS by offset

MikeZeDev:
When pasting a link with search parameters, Common.FetchMangaCSS lose the parameters on the way. Its breaking some plugins that works otherwise.
Ex : AGCTranslation > Arifureta > http://www.agcscanlation.it/progetto.php?nome=Arifureta2. Identifier become /progetto.php instead of  /progetto.php?nome=Arifureta2.

Use case for MangasMultiPageCSSByOffset : AllHentai but its a generic thing that may come handy i guess so i put it in common
https://2023.allhen.online/list?offset=70 , https://2023.allhen.online/list?offset=140, etc...

### DIFF
--- a/src/engine/websites/DoujinDesu.ts
+++ b/src/engine/websites/DoujinDesu.ts
@@ -4,21 +4,11 @@ import { DecoratableMangaScraper } from '../providers/MangaPlugin';
 import * as MangaStream from './decorators/WordPressMangaStream';
 import * as Common from './decorators/Common';
 
-const queryPages = 'div.main div img[src]:not([src=""])';
-
-const script = `
-      new Promise((resolve, reject) => {
-          setTimeout(() => {
-              const images = [...document.querySelectorAll('${queryPages}')];
-              resolve(images.map(image => image.dataset['lazySrc'] || image.dataset['src'] || image.getAttribute('original') ||  image.src));
-          }, 2500);
-      });
- `;
-
-@Common.MangaCSS(/^https?:\/\/212\.32\.226\.234\/manga\/[^/]+\/$/, 'section.metadata h1.title', Common.ElementLabelExtractor('span.alter'))
-@Common.MangasMultiPageCSS('/manga/page/{page}/', 'div.entries article.entry a', 1, 1, 0, Common.AnchorInfoExtractor(true))
+const extractor = Common.AnchorInfoExtractor(true);
+@MangaStream.MangaCSS(/^https?:\/\/212\.32\.226\.234\/manga\/[^/]+\/$/, 'div#infoarea div.post-body h1.entry-title')
+@Common.MangasMultiPageCSS('/komik-list/page/{page}/', '#main .relat div.animepost a', 1, 1, 0, extractor)
 @MangaStream.ChaptersSinglePageCSS('div#chapter_list div.epsleft span.lchx a')
-@Common.PagesSinglePageJS(script)
+@MangaStream.PagesSinglePageCSS()
 @Common.ImageDirect()
 export default class extends DecoratableMangaScraper {
 
--- a/src/engine/websites/DoujinDesu.ts
+++ b/src/engine/websites/DoujinDesu.ts
@@ -4,11 +4,21 @@ import { DecoratableMangaScraper } from '../providers/MangaPlugin';
 import * as MangaStream from './decorators/WordPressMangaStream';
 import * as Common from './decorators/Common';
 
-const extractor = Common.AnchorInfoExtractor(true);
-@MangaStream.MangaCSS(/^https?:\/\/212\.32\.226\.234\/manga\/[^/]+\/$/, 'div#infoarea div.post-body h1.entry-title')
-@Common.MangasMultiPageCSS('/komik-list/page/{page}/', '#main .relat div.animepost a', 1, 0, extractor)
+const queryPages = 'div.main div img[src]:not([src=""])';
+
+const script = `
+      new Promise((resolve, reject) => {
+          setTimeout(() => {
+              const images = [...document.querySelectorAll('${queryPages}')];
+              resolve(images.map(image => image.dataset['lazySrc'] || image.dataset['src'] || image.getAttribute('original') ||  image.src));
+          }, 2500);
+      });
+ `;
+
+@Common.MangaCSS(/^https?:\/\/212\.32\.226\.234\/manga\/[^/]+\/$/, 'section.metadata h1.title', Common.ElementLabelExtractor('span.alter'))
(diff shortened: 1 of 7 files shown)


ronny1982:
Instead of introducing a new extension method and decorator method, i would suggest to add an additional parameter to `FetchMangasMultiPageCSS` and update its implementation since they are very similar e.g.,
```typescript
export async function FetchMangasMultiPageCSS<E extends HTMLElement>(this: MangaScraper, provider: MangaPlugin, path: string, query: string, start = 1, step = 1, throttle = 0, extract = DefaultInfoExtractor as InfoExtractor<E>): Promise<Manga[]> {
    // ...
    for(let offset = start, run = true; run; offset += step) {
        // ...
        const mangas = await FetchMangasSinglePageCSS.call(this, provider, path.replace('{offset}', `${offset}`), query, extract as InfoExtractor<HTMLElement>);
        // ...
    }
    // ...
}
```

Such change would affects many existing sites

src/engine/websites/decorators/Common.ts
```diff
@@ -83,7 +83,7 @@ export async function FetchMangaCSS(this: MangaScraper, provider: MangaPlugin, u
     const uri = new URL(url);
     const request = new FetchRequest(uri.href);
     const data = (await FetchCSS<HTMLElement>(request, query)).shift();
-    return new Manga(this, provider, uri.pathname, extract.call(this, data));
+    return new Manga(this, provider, uri.pathname + uri.search, extract.call(this, data));
```
ronny1982, line 86: The usage of `pathname` was by intention as `search` or `hash` components of an URL makes no sense to be used for resource identification purposes.
I guess some websites don't care about the correct usage of URL components.

Should `uri.hash` also be added?
Don't forget to update [document tag](https://github.com/manga-download/haruneko/blob/1c86efa10a212f20dd8d7406b836f62d60a19876/src/engine/websites/decorators/Common.ts#L75) which explicitly states that only the `pathname` is used for identification.

MikeZeDev: I dont think its necessary to add uri.hash. I mean, uri.hash is barely used for resource identification i've only seen in Mega.nz url to separate the key from the url. But since  i dont see any bad consequences doing so we may as well add it.

As for the search for identifying a resource, thats just the old php way and we should keep it for legacy websites.

Ill modify the informations accordingly

MikeZeDev: I made uri.search & uri.hash inclusion conditional with 2 new optional parameters. This should not affect websites at all.

MikeZeDev:
Changed Common. accordingly. Now its surprising but not so many websites are affected. Many use Common.MangasMultiPageCSS and Common.FetchMangasMultiPageCSS without enough arguments to be affected.

Affected list : 

* Madara decorator
* AnyACG decorator
* MangaNel decorator
* DoujinDesu
* MangaFox

Gotta update them asap

ronny1982:
Make sure to update all affected consumers of `Common.MangasMultiPageCSS` such as
- DoujinDesu
- MangaFox
- AnyACG
- FoolSlide
- MangaNel
- WordPressMadara
- ...

src/engine/websites/DoujinDesu_e2e.ts
```diff
@@ -1,25 +1,25 @@
...
+        title: 'DoujinDesu'
+    },
+    container: {
+        url: 'https://212.32.226.234/manga/hero-villain/',
+        id: '/manga/hero-villain/',
+        title: 'Hero Villain'
+    },
+    child: {
+        id: '/2022/02/19/hero-villain-chapter-01/',
+        title: 'Hero Villain Chapter 01'
```
ronny1982, line 15: I would expect that the manga title is stripped of the chapter names

src/engine/websites/DoujinDesu_e2e.ts
```diff
@@ -1,25 +1,25 @@
-import { TestFixture, type Config } from '../../../test/WebsitesFixture';
```
ronny1982, line 1: The whole file is marked as changed, could it be that it was saved with wrong encoding (non-UTF-8)

src/engine/websites/DoujinDesu.ts
```diff
@@ -1,22 +1,32 @@
-import { Tags } from '../Tags';
```
ronny1982, line 1: The whole content of this file is marked as changed, maybe saved in different encoding?

I suggest to not fix the website functionality itself within this PR, as it is a different scope => separate PR.
Only update the new parameter for `Common.MangasMultiPageCSS`.

In case of a follow-up PR for DoujinDesu i recommend to get the images from the `/themes/ajax/ch.php` endpoint
```sh
curl 'https://212.32.226.234/themes/ajax/ch.php' \
  -H 'content-type: application/x-www-form-urlencoded; charset=UTF-8' \
  -H 'cookie: PHPSESSID=qtbaqp0k6tcnc97bp3m5824hp7' \
  -H 'origin: https://212.32.226.234' \
  -H 'referer: https://212.32.226.234/ikusa-otome-to-ikusa-goto-onna-samurai-hen-part3/' \
  -H 'x-requested-with: XMLHttpRequest' \
  --data-raw 'id=21309' 
```

src/engine/websites/DoujinDesu.ts
```diff
@@ -1,22 +1,32 @@
...
+              const images = [...document.querySelectorAll('${queryPages}')];
+              resolve(images.map(image => image.dataset['lazySrc'] || image.dataset['src'] || image.getAttribute('original') ||  image.src));
+          }, 2500);
+      });
+ `;
+
+@Common.MangaCSS(/^https?:\/\/212\.32\.226\.234\/manga\/[^/]+\/$/, 'section.metadata h1.title', Common.ElementLabelExtractor('span.alter'))
+@Common.MangasMultiPageCSS('/manga/page/{page}/', 'div.entries article.entry a', 1, 1, 0, Common.AnchorInfoExtractor(true))
+@MangaStream.ChaptersSinglePageCSS('div#chapter_list div.epsleft span.lchx a')
+@Common.PagesSinglePageJS(script)
```
ronny1982, line 21: Use `Common.PagesSinglePageJS` which has default delay of 2500ms for injected scripts, so the scripted timeout may not even be needed

src/engine/websites/DoujinDesu_e2e.ts
MikeZeDev: Didnt knew what went wrong i use the same editor than usual

src/engine/websites/DoujinDesu.ts
MikeZeDev: ok

PR #65, opened 2023-01-17
Madtheme 100 Chapters limit fixes

MikeZeDev:
Proposed solution (reuse code and scrap chapters from the page DOM) doesnt work because the page DOM is limited to 100 chapters unless the user click on a button on the page. The correct way is to make a single AJAX call to the api.

So i come with a decorator again, sorry. I also put the script for pages inside it, while we are at it, since it would more convenient tu update one single script

I didnt change the other decorations its pointless.

Ofc, i feel like decorator could be simplified again.

### DIFF
--- a/src/engine/websites/decorators/MadTheme.ts
+++ b/src/engine/websites/decorators/MadTheme.ts
@@ -1,5 +1,5 @@
 import { FetchRequest, FetchCSS } from '../../FetchProvider';
-import { type MangaScraper, type DecoratableMangaScraper, type Manga, Chapter, type Page } from '../../providers/MangaPlugin';
+import { type MangaScraper, type Manga, Chapter, type Page } from '../../providers/MangaPlugin';
 import * as Common from './Common';
 
 const scriptImageLinks = `
@@ -9,9 +9,6 @@ const scriptImageLinks = `
     });
 `;
 
-/* eslint-disable-next-line @typescript-eslint/no-explicit-any */ //=> A mixin class must have a constructor with a single rest parameter of type 'any[]'
-export type Constructor = new (...args: any[]) => DecoratableMangaScraper;
-
 const queryChapterListBloat = '.chapter-update';
 const queryChapterListLinks = 'ul.chapter-list li a';
 const DefaultInfoExtractor = Common.AnchorInfoExtractor(false, queryChapterListBloat);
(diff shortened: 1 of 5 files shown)


src/engine/websites/decorators/MadTheme.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const scriptImageLinks = `
+    new Promise(resolve => {
+        const images = window.chapImages.split(',');
+        resolve(images.map(image => window.mainServer + image));
+    });
+`;
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */ //=> A mixin class must have a constructor with a single rest parameter of type 'any[]'
+export type Constructor = new (...args: any[]) => DecoratableMangaScraper;
```
ronny1982, line 13: Is the export necessary, i didn't see any imports for this type?
Is this type definition necessary, can the type `Common.Constructor` be used instead?

src/engine/websites/decorators/MadTheme.ts
```diff
@@ -0,0 +1,80 @@
...
+ * An extension method for extracting all chapters for the given {@link manga} using the given CSS {@link query}.
+ * This method utilizes the HTML pages provided by the **MadTheme Admin AJAX endpoint** to extract the chapters.
+ * @param this - A reference to the {@link MangaScraper} instance which will be used as context for this method
+ * @param manga - A reference to the {@link Manga} which shall be assigned as parent for the extracted chapters
+ * @param query - A CSS query to locate the elements from which the chapter identifier and title shall be extracted
+ * @param extract - The extractor to use
+ */
+export async function FetchChaptersSinglePageAJAX(this: MangaScraper, manga: Manga, query = queryChapterListLinks, extract = DefaultInfoExtractor): Promise<Chapter[]> {
+    const id = manga.Identifier.split('/').pop();
+    const uri = new URL('/api/manga/' +id+ '/chapters?source=detail', this.URI);
```
ronny1982, line 57: Personally i prefer string interpolation when inserting variable values in the middle of a string literal
```javascript
`/api/manga/${id}/chapters`
```

ronny1982 (approved):
Looks good. I added some subjective suggestions which may or may not be addressed.
Breaking tests seems unrelated to this PR and therefore don't pose a problem.

src/engine/websites/decorators/MadTheme.ts
```diff
@@ -0,0 +1,80 @@
...
+ *************************************************/
+
+/**
+ * A class decorator that adds the ability to extract all pages for a given chapter using the given JS {@link script}.
+ * The pages are extracted from the composed url based on the `Identifier` of the chapter and the `URI` of the website.
+ * @param script - A JS script to extract the image links
+ * @param delay - An initial delay [ms] before the {@link script} is executed
+ */
+export function PagesSinglePageJS(script = scriptImageLinks, delay = 0) {
+    return function DecorateClass<T extends Constructor>(ctor: T): T {
```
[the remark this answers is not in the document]
MikeZeDev: Sure

src/engine/websites/decorators/MadTheme.ts
MikeZeDev: Ill fix that

PR #77, opened 2023-01-22
Properly remove/detect Referer Headers

### LINKED ISSUE #74: Referer issues
1) Requests are sent with default referer aka http://localhost:3000 

Default behavior should be at least sending truncated uri from the request itself.

2) I cant seems to use a custom referer.

Ex : CartoonMad : Chapter list is https://www.cartoonmad.com/comic/1152.html.
To fetch one chapter html page i need to send the chapter list url as referer. Which i cant.

ronny1982:
Resolves #74


MikeZeDev (approved):
* Default requests got a referer
* Changing a FetchRequest referer worked

LGTM, thanks

Issue #98, opened 2023-02-19
Set-Cookie in response issue


MikeZeDev:
Case: 

URL 1 : https://mangayeh.com/chapter/cheat-kusushi-no-slow-life-isekai-ni-tsukurou-drugstore/chapter-44

This url redirect (302) to

URL 2 :  https://gamesk.pro/chapter/cheat-kusushi-no-slow-life-isekai-ni-tsukurou-drugstore/chapter-44?t=db2ef9772b744e93ac21d023d74e6446e1612014959. 

Then there is another redirection on URL2 (302 again) which is 

URL3 : https://gamesk.pro/games/d/serial-cleaners-459051 (randomized)

In browser, URL3 behave differently if you open it in a new tab than if you followed redirections.

Direct access : ads for videogames
Redirected access  :  page with chapters pictures.

![image](https://user-images.githubusercontent.com/10671518/219958404-f01ef1a2-6069-492b-86e6-47eb92d1fe41.png)

If the last request, Firefox got JSESSION in cookies, which Haru didnt. Cookie which should have been added according to the 'Set-Cookie' directive in PREVIOUS Request response, even if the response code was 302.

This may definitely be this JSESSION cookie that we miss. That and maybe referer (but i added it and it doesn't change anything) or the cloudflare cookie (that will be a problem). 

In any case, we should have seen JSESSION cookie in last request i guess.

### Solution ; 

Intercept the requests response (with a handler) and set the cookie.

ronny1982:
I [fixed a timeout issue](https://github.com/manga-download/haruneko/commit/8e830d0f2aa46b8abf0d1b0914c2f4f4b36983bd) with `FetchWindowCSS`.
It is expected that cookies are now correctly set and image links can be extracted from a chapter.

```typescript
console.log('Cookies (before):', await chrome.cookies.getAll({}));
const request = new FetchRequest('https://mangayeh.com/chapter/cheat-kusushi-no-slow-life-isekai-ni-tsukurou-drugstore/chapter-44', {});
const data = await FetchWindowCSS<HTMLImageElement>(request, '.container p img[data-src]', 2500);
console.log('Cookies (after):', await chrome.cookies.getAll({}));
console.log('Images:', data.map(img => img.dataset.src));
```

<img width="1047" alt="image" src="https://user-images.githubusercontent.com/27136773/221345317-10666549-1f64-492f-8c0b-77be464f6875.png">

MikeZeDev:
[...]

PR #99, opened 2023-02-19
Handle Set-Cookie Response Header

MikeZeDev:
Fixes https://github.com/manga-download/haruneko/issues/98 (website still bugged)

Its a first attempt there is certainly an issue with chrome.cookies.set.


web/src/engine/FetchProviderNodeWebkit.ts
```diff
@@ -46,6 +46,30 @@ function ModifyRequestHeaders(details: chrome.webRequest.WebRequestHeadersDetail
 }
 
 function ModifyResponseHeaders(details: chrome.webRequest.WebResponseHeadersDetails): chrome.webRequest.BlockingResponse {
+
+    //Handle Set-Cookie header according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
+    const cookiesHeaders = details.responseHeaders.filter(header => header.name.toLowerCase() == 'set-cookie');
+
+    //Multiple Set-Cookie headers can be sent in the same response.
+    if (cookiesHeaders.length > 0) {
```
Sheepux, line 54: no need for this `if` as  `foreach` will iterate only on elements.
filter can be `null` then just use `cookiesHeaders?.forEach`

MikeZeDev:
Testing this for a while, go errors 
image

Maybe we shouldnt remove space, nor use lowercase. I feel like it will be safer to compare the key name to an exhaustive of attribute (path, expires, httponly, etc...) and fill details using the matching key rather than directly using the key from the SetCookie header.

MikeZeDev:
`FetchProviderNodeWebkit.ts:80 Uncaught (in promise) TypeError: Error in invocation of cookies.set(object details, optional function callback): Error at parameter 'details': Error at property 'expirationDate': Invalid type: expected number, found string.`

Damn, i know why but seriously...

ronny1982:
Use `FetchBrowser` if you want cookies from the web server to be set in HakuNeko.
Self-managing cookies should be avoided.

MikeZeDev:
> Use `FetchBrowser` if you want cookies from the web server to be set in HakuNeko. Parsing and self-managing response cookies should be avoided.

FetchBrowser ? I see no FetchBrowser in code.

> May also try to use the Fetch API with option credentials: same-origin/include, maybe this will allow fetch to set cookies automatically

Cookies to be sent arent in the request because they are not set at all on browser storage in the first place. Using includes indeed send everything but there is not the new cookie stored

1) I use FetchRequest(chapterURL)
2) Response is code 302. Request move to other location
3) Following response  302 again and this time there is a Setcookie header  with "JSESSIONID=xxxxxxxx"
4) Last location request doesnt get the cookie from SetHeader according to console because its even missing from cookie storage. Page displayed is garbage .

Meanwhile on Firefox : 
1) Request (chapterURL)
2) Response is code 302. Request move to other location
3) Following response  302 again and this time there is a Setcookie header  with "JSESSIONID=xxxxxxxx"
4) Last location request get the cookie from SetHeader. Page displayed is chapter viewer

Related :  https://stackoverflow.com/questions/61555100/cookie-in-set-cookie-header-not-being-set
https://stackoverflow.com/questions/31058542/set-cookie-header-not-setting-cookie-in-chrome

ronny1982:
Suggestion to abandon this PR for the following reasons:
- The `Fetch API` is usually not supposed to process the `Set-Cookie` headers and HakuNeko should comply with browser behavior as much as possible (otherwise it might be used by websites to detect and block HakuNeko e.g., `'Set-Cookie: HakuNeko=TRUE'`)
- Maintaining and testing the cookie parser to work correctly with all possible permutations of `Set-Cookie` header values is additional work which should be avoided
- The `FetchProvider` module already has methods which will process the `Set-Cookie` headers
  - `FetchWindowCSS` => now fixed
  - `FetchWindowScript`
  - `FetchWindowPreloadScript`)

Issue #113, opened 2023-03-18
Use correct Accept Header mime type when directly fetching image


MikeZeDev:
### Case : https://raw.senmanga.com/akuyaku-reijou-no-shitsuji-sama-hametsu-flag-wa-ore-ga-tsubusasete-itadakimasu

THIS IS NOT A FILE RENAMING PROBLEM BUT A REQUEST PROBLEM.

Images links ends with webp. Browser downloads WEBP. Haruneko downloads files and rename as jpg **as content is really JPG**.

Browser uses

`Accept : text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8`

Haru uses 

`Accept : */*`

### Solution Proposed

In Common.ImageDirect, try to guess mime type of the link using  MangaPlugin.mimeFileExtension dictionary and modify request headers accordingly.

MikeZeDev:
Maybe use a regex instead of "endswith"?

PR #114, opened 2023-03-18
Common.ImageDirect : Accept Header Fix

MikeZeDev:
Fixes https://github.com/manga-download/haruneko/issues/113

### DIFF
--- a/web/src/engine/providers/MangaPlugin.ts
+++ b/web/src/engine/providers/MangaPlugin.ts
@@ -9,7 +9,7 @@ import { MediaContainer, StoreableMediaContainer, MediaItem, MediaScraper } from
 const settingsKeyPrefix = 'plugin.';
 
 // See: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types
-const mimeFileExtension = {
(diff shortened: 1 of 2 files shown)


ronny1982 (commented):
The `Acccept` header seems to be a general problem in NW.js, it always sends `*/*` instead of the corresponding chromium header which is way more detailed.

The solution in this PR is to specific. Suggestion to inspect why NW.js is not sending the correct `Accept` header that would equal the used chrome browser.

If a fix is really necessary within HaruNeko code, [ModifyRequestHeaders](https://github.com/manga-download/haruneko/blob/master/web/src/engine/FetchProviderNodeWebkit.ts#L27-L46) would be a more appropriate place for this.

MikeZeDev:
I see. I had the same problem in Hakuneko with websites using a specific template. Graphql (?) requests werent working unless you modify Accept header.

The thing is, "Accept */*" works everytime except for a few exceptions. I choose this place on purpose.
Depending on how many exceptions we may indeed need to either make specific modification in website plugin, decorator, or ModifyRequestHeaders like you suggested.

PR #116, opened 2023-03-20
Legacy To Haru 14

MikeZeDev:
* RawDevart
* RawSenManga
* ReadM
* ReadManga


MikeZeDev:
RawSenManga test is accurate for now. It will be updated when https://github.com/manga-download/haruneko/pull/114 is merged, as it impact the file send by the CDN (same url, different content)

web/src/engine/websites/RawDevart.ts
```diff
@@ -0,0 +1,98 @@
...
+            headers: {
+                'Referer': this.URI.href,
+                'Origin': this.URI.href,
+                'X-ReadAuth': key,
+            }
+        });
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return /https?:\/\/endevart\.com\/comic\//.test(url);
```
ronny1982, line 58: Regex can be improved as the part after `comic` consists of characters only allowed for a GUID and then must be terminated with `$`

web/src/engine/websites/RawDevart.ts
```diff
@@ -0,0 +1,98 @@
...
+    public constructor() {
+        super('rawdevart', `RawDevart`, 'https://endevart.com', Tags.Language.Multilingual, Tags.Media.Manga, Tags.Media.Manhua, Tags.Media.Manhwa, Tags.Source.Aggregator);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    private generateRequest(url: string, key: string): FetchRequest {
+        return new FetchRequest(url, {
```
ronny1982, line 48: Suggestion to generate key within this method instead of passing as parameter.

PR #117, opened 2023-03-20
UI classic: settings store more svelte like

Sheepux:
Reworked Ronny's early svelte store implementation to feel more "svelte like" on usage
this can be seen here :
https://github.com/manga-download/haruneko/pull/117/files#diff-1b50d9a5d1ae3e40f8cbfdfb386679aa0b238bc033b49e14118622004e55a004L26-R24
where 
`    <Toggle
        toggled={$ContentPanelValue}
        on:toggle={(evt) => (ContentPanel.Value = evt.detail.toggled)}
    />`
becomes
` <Toggle bind:toggled={$ContentPanel} />`

Ended up with 2 helpers, one for the custom settings of the frontend, and one for the engine's ones
https://github.com/manga-download/haruneko/pull/117/files#diff-043205968d154e9ad9ffa00608ef5900128c29a75482eeca43072b9b7cc7db9e

Fusing both Setting and SettingValue lead to a small issues with the Choice kind (to get the available options) that I bypassed by casting (couldn't cast inline because of svelte)
https://github.com/manga-download/haruneko/pull/117/files#diff-1b50d9a5d1ae3e40f8cbfdfb386679aa0b238bc033b49e14118622004e55a004R14-R17

### DIFF
--- a/web/src/frontend/classic/components/settings/SettingsPanel.svelte
+++ b/web/src/frontend/classic/components/settings/SettingsPanel.svelte
@@ -3,89 +3,68 @@
     import SettingItem from './SettingItem.svelte';
     import {
         ContentPanel,
-        ContentPanelValue,
         Key,
         Locale,
         FuzzySearch,
-        FuzzySearchValue,
         Theme,
-        ThemeValue,
         ViewerDoublePage,
-        ViewerDoublePageValue,
         ViewerMode,
-        ViewerModeValue,
         ViewerReverseDirection,
-        ViewerReverseDirectionValue,
     } from '../../stores/Settings';
+    import type { Choice } from '../../../../engine/SettingsManager';
+
+    const ThemeSetting = Theme.setting as Choice;
+    const ViewModeSetting = ViewerMode.setting as Choice;
 </script>
 
 <SettingItem
-    labelText={$Locale[ContentPanel.Label]()}
-    helperText={$Locale[ContentPanel.Description]()}
+    labelText={$Locale[ContentPanel.setting.Label]()}
+    helperText={$Locale[ContentPanel.setting.Description]()}
 >
-    <Toggle
-        toggled={$ContentPanelValue}
-        on:toggle={(evt) => (ContentPanel.Value = evt.detail.toggled)}
-    />
+    <Toggle bind:toggled={$ContentPanel} />
 </SettingItem>
 
 <SettingItem
-    labelText={$Locale[Theme.Label]()}
-    helperText={$Locale[Theme.Description]()}
+    labelText={$Locale[Theme.setting.Label]()}
+    helperText={$Locale[Theme.setting.Description]()}
 >
-    <Select
-        selected={$ThemeValue}
-        on:update={(evt) => (Theme.Value = evt.detail.toString())}
-    >
-        {#each Theme.Options as option}
+    <Select bind:selected={$Theme}>
+        {#each ThemeSetting.Options as option}
             <SelectItem value={option.key} text={$Locale[option.label]()} />
         {/each}
     </Select>
 </SettingItem>
 
 <SettingItem
-    labelText={$Locale[ViewerMode.Label]()}
-    helperText={$Locale[ViewerMode.Description]()}
+    labelText={$Locale[ViewerMode.setting.Label]()}
+    helperText={$Locale[ViewerMode.setting.Description]()}
 >
-    <Select
-        selected={$ViewerModeValue}
-        on:update={(evt) => (ViewerMode.Value = evt.detail.toString())}
-    >
-        {#each ViewerMode.Options as option}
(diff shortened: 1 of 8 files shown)


web/src/frontend/classic/stores/Helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+import type { Key } from '../../../engine/SettingsGlobal';
+
+const globalsettings = HakuNeko.SettingsManager.OpenScope();
+
+/**
+ * Create a writable svelte store that is coupled with the given setting
+ * and updated whenever the value of the underlying setting is changed.
+ * @param setting - a specific setting of the frontend
+ */
+export function CreateWritableStore<T extends IValue>(setting:Setting<T>) {
```
ronny1982, line 12: Method name is a bit misleading as it creates a svelte store, but return a custom object extending a svelte store.
Suggestion to define a type with expressive name for this returned custom object.

Sheepux: How about now ?

ronny1982: Consider to make method generic and add type of the setting e.g., `Create<V extends IValue, S extends Setting<V>>(setting: S)`

web/src/frontend/classic/stores/Helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+    const { subscribe, set } = writable(setting.Default);
+
+    setting.ValueChanged.Subscribe(
+        (_: typeof setting, value: T) => set(value)
+    );
+    return {
+        subscribe,
+        set: (value: T) => { setting.Value = value; set(value); },
+        reset: () => { setting.Value = setting.Default; set(setting.Default); },
+        setting : setting
```
ronny1982, line 28: Based on previous suggestions, explicitly define type as `S`

web/src/frontend/classic/stores/Helpers.ts
ronny1982: Use space after `:` to improve readability of variable name and type declaration

web/src/frontend/classic/stores/Helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+ * and updated whenever the value of the underlying setting is changed.
+ * @param settingKey - an existing key (created in the engine)
+ */
+export function GetEngineSetting<T extends IValue>(settingKey: Key) {
+    const setting: Setting<T> = globalsettings.Get(settingKey);
+    if (!setting) throw new Error(`Setting ${settingKey} does not exists`);
+    return CreateWritableStore<T>(setting);
+}
+
+export function CreateCountStore(initialValue:number, increment:number,minimum = -Infinity, maximum = Infinity) {
```
ronny1982: Use space after `:` to improve readability of variable name and type declaration

web/src/frontend/classic/stores/Helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * Create a writable svelte store that is coupled with the given setting
+ * and updated whenever the value of the underlying setting is changed.
+ * @param settingKey - an existing key (created in the engine)
+ */
+export function GetEngineSetting<T extends IValue>(settingKey: Key) {
+    const setting: Setting<T> = globalsettings.Get(settingKey);
+    if (!setting) throw new Error(`Setting ${settingKey} does not exists`);
```
ronny1982, line 39: Is this error redundant?
TypeScript is expected to fail compilation in case an invalid `Key` is used for `settingKey`.

web/src/frontend/classic/components/settings/SettingsPanel.svelte
```diff
@@ -3,89 +3,68 @@
...
         ViewerDoublePage,
-        ViewerDoublePageValue,
         ViewerMode,
-        ViewerModeValue,
         ViewerReverseDirection,
-        ViewerReverseDirectionValue,
     } from '../../stores/Settings';
+    import type { Choice } from '../../../../engine/SettingsManager';
+
+    const ThemeSetting = Theme.setting as Choice;
```
ronny1982, line 16: Cast will no longer be necessary in case of using the suggested generic approach for `CreateWriteableStore`

web/src/frontend/classic/stores/Helpers.ts
Sheepux: In case the key is correct but the setting couldn't be found in the setting manager (initialisation issue or any other reason) I wanted a way to make sure we see it.
So it was not intended as a compilation check but a runtime check.

web/src/frontend/classic/components/settings/SettingsPanel.svelte
Sheepux: That did the trick !

PR #128, opened 2023-04-07
Update TypeScript and Module Dependencies

ronny1982:
With this PR TypeScript will be updated to `v5`.
Unfortunately this will break the `Fluent Core` frontend, since the `FAST` framework seems not to work yet with the new TypeScript version.

I still suggest to merge this proposed change, so we can start migrating the decorators to the new ECMAScript compatible syntax.

In case a developer is stuck in the `Fluent Core` frontend, it is possible to switch back to the `Classic` frontend via console (F12) with the following command:
```javascript
HakuNeko.SettingsManager.OpenScope().Get('frontend').Value = 'classic';
```


Sheepux (approved):
I'm ok with disabling vite a11y checks for now
Svelte carbon has been working fine on a11y, i've started to move from div on:click to buttons or clickabletiles.
Eg: https://github.com/manga-download/haruneko/commit/3a38678649d55bcdf88fb0937f63cdb82d97ff8d#diff-94d88c3f82b620ec171a13fd47e00ae8cc180e71ee4bddd16a5a08f4b815caabR69
I will try to fix all of those later on.

/LGTM quick regression tests were successfull on svelte interface

PR #129, opened 2023-04-09
Legacy to Haru 16

MikeZeDev:
* SilentSkyScans
* SweetTimeScan
* Tapread
* TBCScans


web/src/engine/websites/SweetTimeScan.ts
```diff
@@ -0,0 +1,22 @@
+import { Tags } from '../Tags';
+import icon from './SweetTimeScan.webp';
+import { DecoratableMangaScraper } from './../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import * as Madara from './decorators/WordPressMadara';
+
+@Madara.MangaCSS(/^https?:\/\/sweetscan\.net\/manga\//, 'div.post-title h1')
```
ronny1982, line 7: The regex will match links that are not mangas [`https://sweetscan.net/manga/a-good-day-to-be-a-dog/cap-70/`](https://sweetscan.net/manga/a-good-day-to-be-a-dog/cap-70/)

web/src/engine/websites/Tapread.ts
```diff
@@ -0,0 +1,98 @@
+import { Tags } from '../Tags';
+import icon from './Tapread.webp';
+import { Chapter, DecoratableMangaScraper, Manga, Page, type MangaPlugin } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import { FetchCSS, FetchJSON, FetchRequest } from './../FetchProvider';
+
+const imgURL = 'https://static.tapread.com';
+
+type APIMangas = {
```
ronny1982, line 9: It seems not all these fields are used in the implementation.
Suggestion to only declare fields that are actually referenced in source code.
Other API type definitions as well...

web/src/engine/websites/Tapread.ts
```diff
@@ -0,0 +1,98 @@
...
+    public constructor() {
+        super('tapread', `Tapread`, 'https://www.tapread.com', Tags.Language.English, Tags.Source.Official, Tags.Media.Manga, Tags.Media.Manhua, Tags.Media.Manhwa);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return /^https?:\/\/www\.tapread\.com\/comic\/detail\//.test(url);
```
ronny1982, line 60: Should books also be supported?
https://www.tapread.com/book/detail/35

web/src/engine/websites/Tapread.ts
```diff
@@ -0,0 +1,98 @@
...
+            const mangas = await this.getMangasFromPage(page, provider);
+            mangas.length > 0 ? mangalist.push(...mangas) : run = false;
+        }
+        return mangalist;
+    }
+
+    private async getMangasFromPage(page: number, provider: MangaPlugin): Promise<Manga[]> {
+        const url = new URL('/comic/moredetail?pageNo=' + page, this.URI).href;
+        const request = new FetchRequest(url);
+        const data = await FetchJSON<APIMangas>(request);
```
ronny1982, line 81: May consider object decomposition `const { code, result: { chapterList }} = await FetchJSON<APIMangas>(request);`

ronny1982 (approved):

web/src/engine/websites/Tapread.ts
MikeZeDev: For now, HaKuneko only supports /comic/ (mangas). Haruneko port is on par with Hakuneko. 
Books are novels, and afaik there is still no novel support in Haru.

Nothing prevent use to add Novel support in the future.

web/src/engine/websites/Tapread.ts
MikeZeDev: Sure but, why and how ?  I am not really familiar with that, sorry.

ronny1982: It may improve readability e.g., when (re-)using multiple nested properties:
```javascript
const data = Fetch<T>();
return data.result.error ? [] : data.result.chapters.map(...);
// or
const { result: { error, chapters } } = Fetch<T>();
return error ? [] : chapters.map(...);
```

MikeZeDev: problem : code is always 200, result can be null (ie page 6 of manga request)

I cant do ```  return result? ....``` and anyway, exception occurs at FetchJSON because there is no "moreDetails" in result..
```

    private async getMangasFromPage(page: number, provider: MangaPlugin): Promise<Manga[]> {
        const url = new URL('/comic/moredetail?pageNo=' + page, this.URI).href;
        const request = new FetchRequest(url);
        const { result: { moreDetailList } } = await FetchJSON<APIMangas>(request);
        return moreDetailList ? moreDetailList.map(element => new Manga(this, provider, String(element.comicId), element.comicName.trim())) : [];
    }
```

PR #133, opened 2023-04-19
Add MangaReaderCMS Decorator

MikeZeDev:


web/src/engine/websites/decorators/MangaReaderCMS.ts
```diff
@@ -0,0 +1,189 @@
+import { FetchRequest, FetchCSS } from '../../FetchProvider';
+import { type MangaScraper, Manga, type Chapter, type Page, type MangaPlugin } from '../../providers/MangaPlugin';
+import * as Common from './Common';
+
+function ChapterInfoExtractor(element : HTMLElement) {
+    const anchor: HTMLAnchorElement = element.nodeName.toLowerCase() === 'a' ? element as HTMLAnchorElement : element.querySelector('a');
```
ronny1982, line 6: Consider to use `element instanceof HTMLAnchorElement` for type detection

ronny1982: Is it safe to assume `anchor` is always set to a valid value or should there be any error handling in case of `undefined`?

web/src/engine/websites/decorators/MangaReaderCMS.ts
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * An extension method for extracting a single manga from the given {@link url} using the given CSS {@link query}.
+ * The `pathname` of the given {@link url} will be used as identifier for the extracted manga.
+ * When the CSS {@link query} matches a `meta` element, the manga title will be extracted from its `content` attribute, otherwise the `textContent` of the element will be used as manga title.
+ * @param this - A reference to the {@link MangaScraper} instance which will be used as context for this method
+ * @param provider - A reference to the {@link MangaPlugin} which shall be assigned as parent for the extracted manga
+ * @param url - The url from which the manga shall be extracted
+ * @param query - A CSS query to locate the element from which the manga title shall be extracted
+ */
+export async function FetchMangaCSS(this: MangaScraper, provider: MangaPlugin, url: string, query: string = queryMangaTitle): Promise<Manga> {
```
ronny1982, line 57: This seems to be just a wrapper for `Common.FetchMangaCSS` with a specific default value for query.
Suggestion to not implement this method and just export `QueryMangaTitle`, so it can be used like this:
```typescript
import * as Common from '../decorators/Common';
import * as MangaReaderCMS from '../decorators/MangaReaderCMS';

export default class extends DecoratableMangaPlugin {
    public override async FetchManga() {
        // ...
        await Common.FetchMangaCSS(..., MangaReaderCMS.QueryMangaTitle);
    }
}
```

ronny1982: The same may be checked for other methods in this class, if these can be easily replaced by methods from `Common` there is no need for an additional implementation

web/src/engine/websites/decorators/MangaReaderCMS.ts
MikeZeDev: 1) OK (i merely pasted the old code )
2) anchor should always be defined for a given website. I think its like this because there maybe case where we have to extract more than the anchor infos so we use the parent control as selector. 

I tested all the (not dead) websites we had, may mangas and it seems consistent.

web/src/engine/websites/decorators/MangaReaderCMS.ts
MikeZeDev: Hum. I'd rather not override fetch methods in websites plugins if i can. I thought the point of decorators would be not to override FetchX methods each time, which increase a bit the volume of code. 

I also think that overriding FetchX methods makes the code a bit harder to read, and we may miss other things added in FetchX in the future (making the code harder to maintain ? )

Moreover doing like you said would not be consistent with already coded decorators (by me or other contributors)

Maybe i can just exports query values, and use them like that ? 

``` @Common.FetchMangaCSS(..., MangaReaderCMS.QueryMangaTitle); ```

Comment (MikeZeDev) [reply web/src/engine/websites/decorators/MangaReaderCMS.ts:57]: Would something like that be ok in a website plugin then? 

``` 
@Common.MangaCSS(/^https?:\/\/manga\.fascans\.com\/manga\/[^/]+$/, MangaReader.queryMangaTitle)
@MangaReader.MangasSinglePageCSS() //unique method so its ok
@Common.ChaptersSinglePageCSS(MangaReader.queryChapters, MangaReader.ChapterInfoExtractor)
@Common.PagesSinglePageCSS(MangaReader.queryPages, MangaReader.ChapterPageExtractor)
@Common.ImageAjax()

```

Comment (ronny1982) [reply web/src/engine/websites/decorators/MangaReaderCMS.ts]: Yes, that meets very much what i had in mind when suggesting the change

Comment (ronny1982) [reply web/src/engine/websites/decorators/MangaReaderCMS.ts]: And also yes, the `MangasMultiPageCSS` methods could be replaced by the `Common` decorator, since they are just a wrapper with a pre-defined path

File: web/src/engine/websites/decorators/MangaReaderCMS.ts [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * A class decorator that adds the ability to extract multiple mangas from a range of pages using the given CSS {@link query}.
+ * The range begins with 1 and is incremented until no more new mangas can be extracted.
+ * This decorator utilizes the HTML pages provided by the **WordPress Admin AJAX endpoint** to extract the mangas.
+ * @param query - A CSS query to locate the elements from which the manga identifier and title shall be extracted
+ * @param path - An additional prefix for the ajax endpoint relative to the scraper's base url
+ */
+export function MangasSinglePageCSS(query = queryMangas, path = pathname) {
```
ronny1982, line 76: Since we have TypeScript 5 now, i would suggest to avoid the experimental (deprecated) decorator implementation and use the ECMAScript compliant new decorator implementation.

Unfortunately there is not much documentation yet available for this right now, so probably needs some investigation.

ronny1982 (approved):

web/src/engine/websites/decorators/MangaReaderCMS.ts
MikeZeDev: Something to keep in mind then. OK

PR #146, opened 2023-05-21
Add ReaderFront decorator & websites

MikeZeDev:
Used by ScyllaScans and RavensScans


web/src/engine/websites/ScyllaScans.ts
```diff
@@ -0,0 +1,24 @@
+import { Tags } from '../Tags';
+import icon from './ScyllaScans.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import * as ReaderFront from './decorators/ReaderFront';
+
+const apiUrl = 'https://api.scyllascans.org';
+
```
ronny1982, line 8: For consistency consider to use an intermediate variable with expressive name as in the other websites
`const cdnUrl = apiUrl;`

web/src/engine/websites/RavensScans.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@ReaderFront.MangaAJAX(/^https?:\/\/ravens-scans\.com\/work\/[a-z]{2}\/[^/]+/, apiUrl)
+@ReaderFront.MangasSinglePageAJAX(apiUrl)
+@ReaderFront.ChaptersSinglePageAJAX(apiUrl)
+@ReaderFront.PagesSinglePageAJAX(apiUrl, cdnUrl)
+@Common.ImageAjax(true)
+
+export default class extends DecoratableMangaScraper {
+    public constructor() {
+        super('ravensscans', `RavensScans`, 'https://ravens-scans.com', Tags.Language.Multilingual, Tags.Source.Scanlator, Tags.Media.Manga);
```
ronny1982, line 18: This is a new website ID, any plans on how bookmark migration shall be done for `ravenscans-en` and `ravenscans-es`?

web/src/engine/websites/decorators/ReaderFront.ts
```diff
@@ -0,0 +1,304 @@
+import { FetchRequest, FetchGraphQL } from '../../FetchProvider';
```
ronny1982, line 1: May add a file comment with additional information:
https://github.com/dvaJi/ReaderFront

web/src/engine/websites/RavensScans.ts
MikeZeDev: Ah. To be fair, i've never heard about a bookmark migration plan.

I made it like that because api supports both language ,we are able to tag medias and there is a common list for EN and ES mangas (not a different domain, no cookie shenanigans for langage, no token, etc..)

I should have kept both?

ronny1982: At some point, users may switch from HakuNeko, therefore a bookmark import is required. So far all websites have kept their ID, so bookmarks will work out of the box. If a website ID is different from HakuNeko, bookmark migration is necessary.

Either keep two separate websites, or create a ticket for RavenScans bookmark migration with details on how the IDs were changed.

PR #148, opened 2023-05-29
MangareaderCMS : add extractor argument

MikeZeDev:


MikeZeDev:
Tested all related plugins, everything is fine.

PR #152, opened 2023-06-04
Add ReadMangalive decorator

MikeZeDev:
* Based on Readmanga.live (ReadManga)
* Used in ReadManga, MintManga, SelfManga
* Try to fetch images from other servers in case of direct link fail, like original code of ReadManga


Pierre-Monier:
Screenshot 2023-11-24 at 22 51 41

Looks mostly good, just AllHentai website is failing to fetch content

MikeZeDev:
>
> Looks mostly good, just AllHentai website is failing to fetch content

That is expected. Pages are not accessible without login. After logging into the website (there is an open webpage button in fluent core interface) it works. Its written in the code and the e2e test is therefore adapted.

PR #168, opened 2023-07-02
Common Decorator : ADD PagesSubPagesCSS

MikeZeDev:
When you have to fetch more than one HTML page to get all the images links, You know, for those websites where each picture is on a different page (often with <select option= page1.html>

* Fetch Chapter page
* gather subpages links
* get each picture link in each subpage

Since i got that on a couple of websites now i think its worth coding a method.


MikeZeDev:
Use cases: All NineMangas sub websites, websites using TAADD template

Issue #180, opened 2023-07-17
BookmarksPlugin : error while loading bookmarks


MikeZeDev:
![image](https://github.com/manga-download/haruneko/assets/10671518/0790acfc-90cf-40d5-8b4b-00e6e706adc7)

I got several bookmarks created while developing SpeedBind decorator, on "Speedbind" branch.
Those are for websites existing in current branch and others are for new websites. All speedbind.

![image](https://github.com/manga-download/haruneko/assets/10671518/023d7fb1-7676-4741-92f9-7669de281c5c)

Now i am on master branch, and the fact that those bookmarks exists prevent the loading of any website plugin :/

Solution : correctly handle invalid bookmarks aka ignore them

PR #181, opened 2023-07-17
BookmarkPlugin : ignore bookmarks with no parent plugin on load

MikeZeDev:
* Website plugin is unavailable for some reason (deleted)
* Bookmarks for it are still here, and it prevents Haruneko to load any plugin.

Fixes https://github.com/manga-download/haruneko/issues/180


ronny1982:
I'm not sure if this solution would satisfy the user. Imagine a user has 100 bookmarks on MangaDex. Now for some reason MangaDex plugin is removed from HakuNeko, the user will no longer be able to see what was bookmarked on this website.

I suggest that this should be correctly handled in the frontend.

MikeZeDev:
Oh my bad, my title is misleading. In fact, with this PR,  invalid bookmarks are not deleted but ignored.
They stay in Bookmarks database.

image

ronny1982:
I suggest to introduce a dedicated MediaContainer for orphaned bookmarks which may have a localised label such as `Website Terminated (${identifier})` and some kind of warning icon ⚠️.
Than there is no need for special handling of `undefined` in various places, as the parent is a valid MediaContainer.

web/src/engine/providers/BookmarkPlugin.ts
```diff
@@ -24,6 +24,7 @@ export class BookmarkPlugin extends MediaContainer<Bookmark> {
 
     private Deserialize(serialized: BookmarkSerialized): Bookmark {
         const parent = this.plugins.WebsitePlugins.find(plugin => plugin.Identifier === serialized.Media.ProviderID);
+        if (!parent) return undefined; //if plugin is not found, invalidate bookmark
```
ronny1982, line 27: I suggest to introduce a dedicated MediaContainer for orphaned bookmarks which may have a localised label such as `Website Terminated (${identifier})` and some kind of warning icon ⚠️.
Than there is no need for special handling of `undefined` in various places, as the parent is a valid MediaContainer.

MikeZeDev: So a MediaItem i would apply as Parent to the bookmark in case MangaPlugin is not found on Load()? Am i right?

MikeZeDev: Tried to do something like that.  For sure it needs improvement but am i on the right track now ?

web/src/engine/providers/BookmarkPlugin.ts
```diff
@@ -23,7 +25,17 @@ export class BookmarkPlugin extends MediaContainer<Bookmark> {
     }
 
     private Deserialize(serialized: BookmarkSerialized): Bookmark {
-        const parent = this.plugins.WebsitePlugins.find(plugin => plugin.Identifier === serialized.Media.ProviderID);
+        let parent = this.plugins.WebsitePlugins.find(plugin => plugin.Identifier === serialized.Media.ProviderID);
+        const dummyPluginName = `deleted_${serialized.Media.ProviderID}`;
+        if (!parent) {
```
ronny1982, line 30: I think you don't have to create a full fledged plugin and also don't have to add it to the plugin list (which would expose all these dummy plugins to the user).

Try to use a more minimalistic approach, if there is no parent, just create an object that satisfies the `IMediaContainer` interface.

```javascript
_createInvalidParent(name: string, identifier, string): IMediaContainer {
    return {
        // dummy properties and empty functions for IMediaContainer
    };
}
// ...
const parent = this.plugins.WebsitePlugins.find(/*...*/) ?? this._createInvalidParent('[DELETED] ' + serialized.Media.Label, serialized.Media.ProviderID);
```

MikeZeDev: Done, i guess. There is an exception on invalid bookmark clicking because Origin doesnt exist but it seems to be handled properly and doesnt impact program usability.

Sheepux:
> Oh my bad, my title is misleading. In fact, with this PR, invalid bookmarks are not deleted but ignored.
> They stay in Bookmarks database.

The bookmark should still be displayed, but we would need an invalid icon and an error message when trying to get the media items.
Not having a parent would be the correct way to detect it in the frontUI.

MikeZeDev:
> Not having a parent would be the correct way to detect it in the frontUI.

But not having a parent causes multiple engine errors. Can't we just define a specific property to true in case of orphaned bookmark, that can be read by front ?

ronny1982:
> The bookmark should still be displayed, but we would need an invalid icon and an error message when trying to get the media items. Not having a parent would be the correct way to detect it in the frontUI.

This is exactly why a stub parent is a good solution, it can provide a ⚠️  icon as well as appropriate handling of the `Update` methods. Furthermore it will provide the original website ID where the bookmark was originally created from, otherwise this information will be lost.

web/src/engine/providers/BookmarkPlugin.ts
```diff
@@ -39,9 +41,65 @@ export class BookmarkPlugin extends MediaContainer<Bookmark> {
         return bookmark;
     }
 
+    _createInvalidParent(name: string, identifier: string): IMediaContainer {
+        const fakePlugin: IMediaContainer = {
+            Identifier: identifier,
+            Title: name,
+            get Entries(): [] {
```
ronny1982, line 48: I think all getters can be replaced by fields
`Entries: []`

web/src/engine/providers/BookmarkPlugin.ts
```diff
@@ -39,9 +41,65 @@ export class BookmarkPlugin extends MediaContainer<Bookmark> {
...
+            },
+            get Icon(): string {
+                return null;
+            },
+
+            get Tags(): Tag[] {
+                return [];
+            },
+
+            IsSameAs(other: IMediaContainer): boolean {
```
ronny1982, line 63: May be simplified to always return `false`

web/src/engine/providers/BookmarkPlugin.ts
```diff
@@ -39,9 +41,65 @@ export class BookmarkPlugin extends MediaContainer<Bookmark> {
...
+                return true;
+            },
+
+            /* eslint-disable-next-line @typescript-eslint/no-unused-vars */ //=> Base class default implementation
+            CreateEntry(_identifier: string, _title: string): IMediaChild {
+                return;
+            },
+
+            /* eslint-disable-next-line @typescript-eslint/no-unused-vars */ //=> Base class default implementation
+            TryGetEntry(_url: string): Promise<IMediaChild> {
```
ronny1982, line 82: May simplify all methods using arrow declaration
`TryGetEntry: (_: string) => { throw new Error(); },`

web/src/engine/providers/BookmarkPlugin.ts
```diff
@@ -39,9 +41,65 @@ export class BookmarkPlugin extends MediaContainer<Bookmark> {
...
+                return;
+            },
+
+            Update(): Promise<void> {
+                return;
+            },
+
+            *[Symbol.iterator](): Iterator<IMediaChild> {
```
ronny1982, line 70: Iterator can be left empty
`[Symbol.iterator]: function* () {},`

web/src/engine/providers/BookmarkPlugin.ts
```diff
@@ -23,7 +25,7 @@ export class BookmarkPlugin extends MediaContainer<Bookmark> {
     }
 
     private Deserialize(serialized: BookmarkSerialized): Bookmark {
-        const parent = this.plugins.WebsitePlugins.find(plugin => plugin.Identifier === serialized.Media.ProviderID);
+        const parent = this.plugins.WebsitePlugins.find(plugin => plugin.Identifier === serialized.Media.ProviderID) ?? this._createInvalidParent('[DELETED] ' + serialized.Media.EntryID, serialized.Media.ProviderID);
```
ronny1982, line 26: Name should probably also contain the ProviderID instead of the MangaID (which can be quite long and may not benefit the user)

Sheepux:
> it can provide a ⚠️ icon a

it's not to the plugin role to provide the UI icon, it's up to the frontend to choose what to do

MikeZeDev:
> > it can provide a ⚠️ icon a
> 
> it's not to the plugin role to provide the UI icon, it's up to the frontend to choose what to do

Do you need some modification in the current PR so you can detect what is a fake parent and then choose to display a warning icon instead of the star icon? :)

Sheepux:
`If parent.id = orphaned then display warning`, i guess that's ok

MikeZeDev:
> `If parent.id = orphaned then display warning`, i guess that's ok

No, more like

```if Bookmark.orphaned then ``` because dummy parent have a proper id so we can identify the missing parent

**Comment by Sheepux** [comment]
If it has a proper id ... then i can identify It ....

**Comment by ronny1982** [comment]
Additional fields or magic values can be avoided, a bookmark is orphaned when its parent is not in the list of website plugins.

```javascript
// Either as external method or as getter on BookmarkManager
isOrphaned(bookmark) {
    return !HakuNeko.Plugins.WebsitePlugins.some(plugin => plugin.IsSameAs(bookmark.Parent));
}
```

MikeZeDev:
Done

ronny1982 (approved):

web/src/engine/providers/BookmarkPlugin.ts
```diff
@@ -36,12 +36,43 @@ export class BookmarkPlugin extends MediaContainer<Bookmark> {
             serialized.Info.EntryID
         );
         bookmark.Changed.Subscribe(this.OnBookmarkChangedCallback.bind(this));
+        console.log('is orphaned : '+ this.isOrphaned(bookmark));
```
ronny1982, line 39: Debug logging should be removed from production code.

MikeZeDev: /facepalm

PR #210, opened 2023-09-12
Olympus : Madara to custom API

MikeZeDev:


web/src/engine/websites/OlympusScanlation.ts
```diff
@@ -1,36 +1,114 @@
...
+        const chapterslist = [];
+        for (let page = 1, run = true; run; page++) {
+            const chapters = await this.getChaptersFromPage(page, manga);
+            chapters.length > 0 ? chapterslist.push(...chapters) : run = false;
+        }
+        return chapterslist;
+    }
+
+    private async getChaptersFromPage(page: number, manga: Manga) : Promise<Chapter[]> {
+        const mangaData = JSON.parse(manga.Identifier);
```
ronny1982, line 99: May use [destructured assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)
`const { slug, type } = JSON.parse(manga.Identifier);`

web/src/engine/websites/OlympusScanlation.ts
```diff
@@ -1,36 +1,114 @@
...
+            mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+        }
+        return mangaList;
     }
+
+    private async getMangasFromPage(page: number, provider: MangaPlugin): Promise<Manga[]> {
+        const uri = new URL(`/api/series?page=${page}&direction=asc`, this.apiUrl);
+        const request = new FetchRequest(uri.href);
+        const data = await FetchJSON<APIMangas>(request);
+        return data.data.series.data.map(element => new Manga(this, provider, JSON.stringify({ slug: element.slug, type: element.type }), element.name.trim()));
```
ronny1982, line 86: IS the dynamic `type` really required, or can we just use a constant value instead?

ronny1982 (approved):

web/src/engine/websites/OlympusScanlation.ts
MikeZeDev: its needed for the api call when you fetch pages. IIRC its kinda always the same but during testing i found some others.
I would rather keep the good one than hardcoding one. Less maintenance.

PR #221, opened 2023-09-25
ModeScanlator : fetch pages from zip file

MikeZeDev:
Website code explicitly only deal with avif and svg. Svg are useless filler, we only use the other files as images.


web/src/engine/websites/ModeScanlator.ts
```diff
@@ -1,22 +1,76 @@
...
 
     public constructor() {
         super('modescanlator', `Mode Scanlator`, 'https://modescanlator.com', Tags.Language.Portuguese, Tags.Media.Manga, Tags.Media.Manhua, Tags.Media.Manhwa, Tags.Source.Scanlator);
     }
-
     public override get Icon() {
         return icon;
     }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
```
ronny1982, line 39: The method `FetchPages` is supposed to be lightweight on resources.
If there are many images, this method will run for a long time (progress bar not moving) and may use a lot of memory.

I suggest this method to return each page with the link to the zip archive and the filename which shall be extracted.
Duplicated zip archive downloads are probably handled by the browser cache.

The `FetchImage(page: Page)` method needs to be implemented accordingly to download he image from the zip archive.

MikeZeDev: I knew you would have said that. My goal was to avoid redownloading the zips for each Page. Since i have the zip buffers in memory how  do you suggest that FetchImage can properly use them ? I feel like page parameters is a bad idea but idk.

ronny1982: Just re-download for now. Keeping HaruNeko's intended concepts is more important right now than optimizing bandwidth usage.

It could be possible that the archive hosted by CloudFlare supports ranged requests, so in the future it might be possible to only extract the bytes from the zip archive needed for a certain image (after reading the zip index).

MikeZeDev: Turns out zips are not redownloaded 👍 
image

web/src/engine/websites/ModeScanlator.ts
```diff
@@ -68,9 +65,20 @@ export default class extends DecoratableMangaScraper {
         }
 
     }
-
     extractNumber(fileName) : number {
         return parseInt(fileName.split(".")[0]);
     }
 
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        if (page.Link.href.endsWith('.zip')) {
```
ronny1982, line 73: No need to await the return statements, the Promises can be returned directly

ronny1982 (approved):

PR #222, opened 2023-09-25
Implement Protobuf 

MikeZeDev:


web/src/engine/websites/legacy/ComicZerosum.ts
```diff
@@ -1,56 +1,94 @@
...
+type APIPages = {
+    pages: {
+        imageUrl: string
+    }[]
+}
 
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+    private api_url = 'https://api.zerosumonline.com/api/v1/';
+    private protoTypes = '/src/engine/websites/legacy/ComicZerosum.proto';
```
ronny1982, line 41: I don't think this path will be available when bundled and deployed.
The proto types probably needs to be bundled as well...

web/src/engine/websites/legacy/ComicZerosum.ts
```diff
@@ -1,56 +1,94 @@
...
     }
 
-    async _getMangas() {
-        const request = new Request(`${this.data_url}/list/name.json`, this.requestOptions);
-        const data = await this.fetchJSON(request);
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        //'https://api.zerosumonline.com/api/v1/title?tag=
+        const mangaid = url.match(/\/detail\/([\w]+)/)[1];
+        const requri = new URL(`title?tag=${mangaid}`, this.api_url);
+        const responseType = 'ComicZerosum.TitleView';
```
ronny1982, line 57: May remove once used intermediate variable

web/src/engine/websites/legacy/ComicZerosum.ts
```diff
@@ -1,56 +1,94 @@
...
+        return data.titles.map(element => {
+            return new Manga(this, provider, element.tag, element.name.trim());
         });
     }
 
-    async _getChapters(manga) {
-        const request = new Request(`${this.data_url}/works/${manga.id}.json`, this.requestOptions);
-        const data = await this.fetchJSON(request);
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        //'https://api.zerosumonline.com/api/v1/title?tag=
```
ronny1982, line 74: Remove redundant comment

web/src/engine/websites/legacy/ComicZerosum.ts
```diff
@@ -1,56 +1,94 @@
...
-        this.data_url = 'https://online.ichijinsha.co.jp/json/zerosum';
+    public override ValidateMangaURL(url: string): boolean {
+        return /https?:\/\/zerosumonline\.com\/detail\//.test(url);
     }
 
-    async _getMangas() {
-        const request = new Request(`${this.data_url}/list/name.json`, this.requestOptions);
-        const data = await this.fetchJSON(request);
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        //'https://api.zerosumonline.com/api/v1/title?tag=
```
ronny1982, line 54: Remove redundant comment

web/src/engine/websites/legacy/ComicZerosum.ts
```diff
@@ -1,56 +1,94 @@
...
+        });
+    }
 
-        return data.Work.Stories.map(story => {
-            return {
-                id: story.Url,
-                title: story.Name
-            };
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        //https://api.zerosumonline.com/api/v1/viewer?chapter_id=
```
ronny1982, line 85: Remove redundant comment

ronny1982 (changes requested):
Make sure proto types are correctly bundled and are working in production build before merging PR

web/src/engine/websites/legacy/ComicZerosum.ts
MikeZeDev: I knew that and i asked myself if  we could have done something about the path problem. Oh well, prototypes are strings included in the ts file now. Changed FetchProto accordingly.

web/src/engine/websites/ComicZerosum.ts
```diff
@@ -0,0 +1,205 @@
...
+type APIPages = {
+    pages: {
+        imageUrl: string
+    }[]
+}
+
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+    private api_url = 'https://api.zerosumonline.com/api/v1/';
+    private protoTypes = `
```
ronny1982, line 41: Probably better to extract this as constant string outside of the class, maybe even outside of the file and import the constant string from there.

ronny1982 (approved):

web/src/engine/websites/ComicZerosum.ts
MikeZeDev: Outside of the class : fine.
Outside of the file? you mean not a proto file but a ts file? I wanna make sure its not impacted by _index.ts rebuilding script? 

Couldn't we juste make sure the building process copy the files, and keep proto files as they are? Which help because they are recognized by code editors.

ronny1982: I assume you can use any file extension you want for a typescript file, but `.ts` is the only one that is inferred.

_FooBar.proto_
```typescript
export const protoTypeDefinition: string = `
message A ...
message B ...
`;
```

_FooBar.ts_
```typescript
import { protoTypeDefinition } from './FooBar.proto';
const root = proto.parse(protoTypeDefinition);
```

ronny1982: If you want a really fancy solution you can write a [tagged template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates) method which can directly be used for string literal (templates)

```typescript
export function proto(strings: string[], ...) {
    return proto.parse(strings.join(''))...;
}

const root = proto`
message A ...
message B ...
`;
```

MikeZeDev:
test & prod works but no svelte checks. :/


image

Issue #112, opened 2023-03-18
Duplicate Mangas/chapters ID handling


MikeZeDev:
Case : https://raw.senmanga.com/mou-hatarakitakunaindesu-boukensha-nanka-yamete-yaru-imasara-taiguu-wo-kaerukara-to-onegai-sarete-mo-okotowaridesu-boku-wa-zettai-hatarakimasen

Its not caused by a bad CSS selector in Haru.
There are 2 chapters 3.1, with the same id. 

When selecting the manga it cause errors, spinning stays, and subsequent click on others mangas keeps the spinning thingie, chapters are added BELOW IT 

![image](https://user-images.githubusercontent.com/10671518/226103660-1f513b1c-fe7c-4e9c-9fa9-8daa87e0c2b5.png)

![image](https://user-images.githubusercontent.com/10671518/226103708-b748edc3-aab6-40e8-9f95-868731ff3ad6.png)

MikeZeDev:
I see no reasons why a chapter id would be duplicated. Different language but same ID? Website must be able to differentiate what to load therefore Haru plugin should compose an identifier using id and language to make it unique anyway.

Would if be doable to filter dupes by default?

Sheepux:
Oh interresting unit test that i should add in SAM
Is that an engine issue or an UI issue ? (going to test)

MikeZeDev:
Error is throwed in node_modules/svelte/internal/index.mjs. 

>  Is that an engine issue or an UI issue ? (going to test)

Both i guess. Error occurs because of engine dont filter duplicate chapters, but UI should have be cleaned after we switched manga.

Sheepux:
That's because i'm using a unique key identifier:
https://github.com/manga-download/haruneko/blob/master/web/src/frontend/classic/components/MediaItemSelect.svelte#L244
`            {#each filteredItems as item (item.Identifier)}`
looks like that having the same item.Identifier is a big issue for svelte (seems logicical tho)

MikeZeDev:
You're not wrong, its the website that is wrong ofc. Nonetheless, maybe something can be done to clean MedialistItem list UI (or whatever you call it) when changing manga (like, force hidding spinner??)

Sheepux:
The only way to not get the internal svelte to crash is to dedupe the list.
This is something that MUST be dore by the engine.

ronny1982:
This is an easy fix, but there are some open questions:
- If the website shows the same chapter twice, should HaruNeko do the same?
[...]

PR #228, opened 2023-10-02
Filter Duplicates mangas & chapters

MikeZeDev:
* Same ID = duplicate It should be ok for most websites

* In case the same identifier (URI, or internal website ID) would lead to different ressources, (ie lang set by cookie) its up to the plugin to create unique ID before filtering occurs (right after FetchMangas / FetchChapters).

* In any case people might override FilterMangas or FilterChapters

Fixes https://github.com/manga-download/haruneko/issues/112


web/src/engine/providers/MangaPlugin.ts
```diff
@@ -76,6 +79,20 @@ export class DecoratableMangaScraper extends MangaScraper {
...
+    public async FilterChapters(list: Chapter[]): Promise<Chapter[]> {
+        return list.filter(
+            (obj, index) =>
+                list.findIndex((item) => item.Identifier === obj.Identifier) === index
+        );
+    }
+
+    public async FilterMangas(list: Manga[]): Promise<Manga[]> {
+        return list.filter(
+            (obj, index) =>
```
MikeZeDev, line 92: One my want to simplfy this with generic types but i dont see how

```
    public async FilterChapters(list: Chapter[]): Promise<Chapter[]> {
        return this.FilterContent<Chapter>(list);

    }

    public async FilterMangas(list: Manga[]): Promise<Manga[]> {
        return this.FilterContent<Manga>(list);

    }

    private async FilterContent<T> (list : T[]) : Promise<T[]>{
        return list.filter(
            (obj, index) =>
                list.findIndex((item) => item.Identifier === obj.Identifier) === index
        );
    } 
```

web/src/engine/providers/MangaPlugin.ts
```diff
@@ -76,6 +79,20 @@ export class DecoratableMangaScraper extends MangaScraper {
     public FetchImage(_page: Page, _priority: Priority, _signal: AbortSignal): Promise<Blob> {
         throw new Error();
     }
+
+    public async FilterChapters(list: Chapter[]): Promise<Chapter[]> {
+        return list.filter(
+            (obj, index) =>
+                list.findIndex((item) => item.Identifier === obj.Identifier) === index
```
ronny1982, line 86: This will just take the first manga with the duplicated identifier in the list, it does not cover the case that the first manga may have the title 'Error' and the second manga may have the title 'My little Sheep'. Obviously, in this case the second manga is the one that should be picked.

web/src/engine/providers/MangaPlugin.ts
ronny1982: No need for generics, since manga and chapter are implementing the `IMediaContainer` interface which has the `Identifier` property.

ronny1982: An alternative to `filter` is ustilizing a `set`
```typescript
const mangas = [{ id: '1' }, { id: '7' }, { id: '1' }, { id: '8' }, { id: '9' }];
const unique = [...new Set(mangas.map(manga => manga.id))].map(id => mangas.find(manga => manga.id === id));
```

ronny1982 (commented):
Is there really a use case that the scraper needs to provide customized methods for filtering?
Couldn't there just be a more central method for `IMediaItems`, that will also be available for animes and novels in the future. Maybe a unique media list transformer similar to the `ImageLinkDeProxifier`.

A website that still needs customized filtering can still do this by overrideing their `FetchMangas` and `FetchChapters` method.

web/src/engine/providers/MangaPlugin.ts
ronny1982:
```typescript
export function GetUniqueMediaItems<T extends IMediaItem>(items: Array<T>): Array<T> {
    return [...items]; // TODO: Only get unique entries by `Identifier` property
}

MikeZeDev: (its chapters here but i see your point)

> Obviously, in this case the second manga is the one that should be picked.

Obviously, but would that "rule" you made about the title be the same everywhere? Also, chapter title may have "Error" in the title just because well, its part of the real title.

The point of this filtering is to be generic, and will be fine 95% of the time. Most of the time it wont be necessary, and when its necessary chapters tends to be named with the same name. I dont think we will find another rule that wont cause a problem on one website because its needed on another.  If you have more idea about filtering rules i am open to discussions ;)

For such edges cases of and edge cases, dev can override the Filter methods 🤷 

The other solution is to filter identical ID + identical title in this, but then the UI would have to be reworked to work with ID+title so there is no bug with only identical ID

MikeZeDev:
> Is there really a use case that the scraper needs to provide customized methods for filtering? Couldn't there just be a more central method to get unique `IMediaItem[]`, that will also be available for animes and novels in the future. Maybe a unique media list transformer similar to the `ImageLinkDeProxifier`. e.g.
> 
> ```ts
> export function GetUniqueMediaItems<T extends IMediaItem>(items: Array<T>): Array<T> {
>     return [...items]; // TODO: Only get unique entries based on `Identifier` property
> }
> ```
> 
> A website that still needs customized filtering can do this by overrideing their `FetchMangas` and `FetchChapters` method.

My point was for the plugin developer to be able to override dupe filtering too, so i put it in scraper, but maybe at that time and right now i am just brainfarting. I am a bit lost with all those IMediaContainer, IMediaChild, IMediaItem, you see i dont crealy grasp the differences and meanings.

Ofc its better to get a generic deduping method, but i am a bit lost.

web/src/engine/providers/MangaPlugin.ts
MikeZeDev: I wondered how to properly use Set for this <3

web/src/engine/providers/MangaPlugin.ts
ronny1982: I have no suggestion how to determine the best fitting of the duplicates.

Another solution that comes to mind would be to keep the duplicates. the object instances itself are different, even if the ID property is the same. So programatically they can be distinguished. If the UI needs a unique identifier, maybe adding a hash method or a symbol field to identify the object instance may solve the problem...

ronny1982:
> > Is there really a use case that the scraper needs to provide customized methods for filtering? Couldn't there just be a more central method to get unique `IMediaItem[]`, that will also be available for animes and novels in the future. Maybe a unique media list transformer similar to the `ImageLinkDeProxifier`. e.g.
> > ```ts
> > export function GetUniqueMediaItems<T extends IMediaItem>(items: Array<T>): Array<T> {
> >     return [...items]; // TODO: Only get unique entries based on `Identifier` property
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > A website that still needs customized filtering can do this by overrideing their `FetchMangas` and `FetchChapters` method.
> 
> My point was for the plugin developer to be able to override dupe filtering too, so i put it in scraper, but maybe at that time and right now i am just brainfarting. I am a bit lost with all those IMediaContainer, IMediaChild, IMediaItem, you see i dont crealy grasp the differences and meanings.
> 
> Ofc its better to get a generic deduping method, but i am a bit lost. EDIT : Ok i think i got it.

Maybe we can remove some redundant interfaces in the future. Coming from C# it is common practice to separate declaration and implementation, but TypeScript supports to infer the declaration from the implementation which may be sufficient in some cases and makes some interfaces obsolete...

web/src/engine/providers/MangaPlugin.ts
MikeZeDev: The hash solution is elegant. Generating one is not a big deal, same for storage but its putting more work on the ui (i mean it need reworking )

MikeZeDev: Ok, I think i'll change the filter to (same id + same title) its already a simple solution worth to be validated which wont hinder further fixes. I'll let the issue opened.

MikeZeDev: Hum, should i change the filtering for (same iD+ same title)? And hope UI can filter dupes using instance?

PR #229, opened 2023-10-03
BacaKomik : fix domain & pictures download

MikeZeDev:
their image hoster hate referrer

### DIFF
--- a/web/src/engine/websites/BacaKomik.ts
+++ b/web/src/engine/websites/BacaKomik.ts
@@ -1,21 +1,31 @@
 import { Tags } from '../Tags';
 import icon from './BacaKomik.webp';
-import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import { DecoratableMangaScraper, type Page } from '../providers/MangaPlugin';
 import * as MangaStream from './decorators/WordPressMangaStream';
-import * as Common from './decorators/Common';
+import type { Priority } from '../taskpool/TaskPool';
+import { Fetch, FetchRequest } from '../FetchProvider';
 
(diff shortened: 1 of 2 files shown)


ronny1982:
I think you stumbled over an interesting finding.
Maybe the method [FetchImage](https://github.com/manga-download/haruneko/blob/c32114e05fdf65dc139b798fb829e3641beab4b7/web/src/engine/websites/decorators/Common.ts#L468) should be changed, so when `pretendImageElementSource` is provided as `true`, no Referer header should be added.

With this change it should be possible to use decorator `@Common.ImageElement(...)` which tries to simulate an image is loaded via `` tag from within a browser.

In case `FetchImage` is updated in the scope of this PR, it would be beneficial to test all connectors currently using the `@Common.ImageElement(...)` decorator to prevent regression bugs.

ronny1982 (approved):

MikeZeDev:
> I think you stumbled over an interesting finding. Maybe the method [FetchImage](https://github.com/manga-download/haruneko/blob/c32114e05fdf65dc139b798fb829e3641beab4b7/web/src/engine/websites/decorators/Common.ts#L468) should be changed, so when `pretendImageElementSource` is provided as `true`, no Referer header should be added.
> 
> With this change it should be possible to use decorator `@Common.ImageElement(...)` which tries to simulate an image is loaded via `` tag from within a browser.
> 
> In case `FetchImage` is updated in the scope of this PR, it would be beneficial to test all connectors currently using the `@Common.ImageElement(...)` decorator to prevent regression bugs.

Actually there is this to consider too : https://www.w3schools.com/TAGS/att_img_referrepolicy.asp

PR #236, opened 2023-10-07
unionmangas : change domain

MikeZeDev:
Using blacklist didnt help so i Initialize without javascript


web/src/engine/websites/UnionMangas.ts
```diff
@@ -3,25 +3,31 @@ import icon from './UnionMangas.webp';
...
+        super('unionmangas', `UnionMangas`, 'https://unionmangasbr.top', Tags.Language.Portuguese, Tags.Media.Manga, Tags.Source.Aggregator);
     }
 
     public override get Icon() {
         return icon;
     }
 
+    //override Initialize because website is so damn slow, probably because ads and tracking
+    public override async Initialize(): Promise<void> {
+        await Fetch(new FetchRequest(this.URI.href));
```
ronny1982, line 25: It seems this `fetch` does not have any use, so it should be removed, just leave the function body empty

ronny1982 (approved):

PR #238, opened 2023-10-09
Moving AssortedScans & ComicK from legacy

MikeZeDev:


web/src/engine/websites/AssortedScans.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const request = new FetchRequest(new URL(chapter.Identifier+'1/', this.URI).href);
+        const data = await FetchCSS<HTMLAnchorElement>(request, 'li.dropdown-element.page-details a');
+        return data.map(page => new Page(this, chapter, new URL(page.pathname, this.URI)));
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        const fakechapter = new Chapter(this, page.Parent.Parent as Manga, page.Link.href, '');
+        const image = (await Common.FetchPagesSinglePageCSS.call(this, fakechapter, '#page-image'))[0];
```
ronny1982, line 30: This request is not part of the rate limiting mechanism, is this by intention?
If it should be rate limited consider to put it on the same rate limit pool as the image download.
```typescript
const image = await this.imageTaskPool.Add(async () => {
    const request = new FetchRequest(page.Link.href, {
        signal: signal
    });
    const image = (await FetchCSS(request, '#page-image'))[0];
    return DeProxify(image.Link).href;
}, priority, signal);
```

ronny1982 (approved):

web/src/engine/websites/AssortedScans.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const request = new FetchRequest(new URL(chapter.Identifier+'1/', this.URI).href);
+        const data = await FetchCSS<HTMLAnchorElement>(request, 'li.dropdown-element.page-details a');
+        return data.map(page => new Page(this, chapter, new URL(page.pathname, this.URI)));
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        const fakechapter = new Chapter(this, page.Parent.Parent as Manga, page.Link.href, '');
```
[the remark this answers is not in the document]
MikeZeDev: FetchCSS is ok

web/src/engine/websites/AssortedScans.ts
MikeZeDev: As always you are right, boss

MikeZeDev: Please tell me if the solution is ok for you.

ronny1982: Looks good.
If you think there might be more websites where the image needs to be extracted from a dedicated HTML page, maybe adding a new `Common` decorator would make sense which fetches the `Page`, extracts the image via CSS query and fetches the image.
`export function ImageCSS(query: string, detectMimeType = false)`

MikeZeDev: I have at least two more; LelScanVF (in another PR ) and Mangatown. That said, mangatown need mangahere referrer to get its pictures :D

PR #245, opened 2023-10-12
Infer the Mock Type for Global Fetch

ronny1982:
Workaround for type incompatibility caused by https://github.com/nodejs/undici/pull/1947
Always use the same type as `global.fetch` for our mock

### DIFF
--- a/web/src/engine/FetchProvider.ts
+++ b/web/src/engine/FetchProvider.ts
@@ -8,10 +8,10 @@ const fail = function() {
     throw new Error();
 };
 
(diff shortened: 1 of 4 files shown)


MikeZeDev:
That fix is indeed correcting the Github syntax check error caused by undici update, as checks are passed.

After forcing modules updates on my end, for diagnosis purposes, i had the error, and with this fix i cant reproduce the error anymore.

Now, i fail to see the consequences of this fix. All i can say is "that works".

MikeZeDev (approved):
Fixes the problem with undici, with unknown consequences (maybe none)

PR #249, opened 2023-10-12
Coreview moving from legacy & add tests

MikeZeDev:
Also fix SundayWebry chapters list

ShonenMagazine & ShonenMagazinePocket have been merged, singe manga links are on the same domain


ronny1982:
If both websites with the same top-level domain are pointing to the same series links, i see no problem in merging both websites.

MikeZeDev:
> If both websites with the same top-level domain are pointing to the same series links, i see no problem in merging both websites.

There is '/series' mangalist path for ShonenMagazinePocket, and '/series/smaga', '/series/bmaga', '/series/others' for  ShonenMagazine. All manga links matches ```/^https?:\/\/pocket\.shonenmagazine\.com\/episode\/\d+$/``` anyway

web/src/engine/websites/ShonenMagazine.ts
```diff
@@ -0,0 +1,38 @@
...
+        super('shonenmagazine', `週刊少年マガジ (Weekly Shonen Magazine & Pocket Magazine)`, 'https://pocket.shonenmagazine.com', Tags.Language.Japanese, Tags.Source.Official, Tags.Media.Manga);
+    }
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+
+        //fetch from pocket.shonenmagazine.com [/series]
+        const mangas = await CoreView.FetchMangasMultiPageCSS.call(this, provider, ['/series'], 'div.series-items ul.daily-series > li.daily-series-item > a');
```
ronny1982, line 25: A shorter, but dirty solution would be to temporary replace the `URI`
```typescript
this.URI.href = 'https://pocket.shonenmagazine.com'
const mangas1 = await CoreView.FetchMangasMultiPageCSS.call(this, provider, ['/series'], 'div.series-items ul.daily-series > li.daily-series-item > a');

this.URI.href = 'https://shonenmagazine.com'
const mangas2 = await CoreView.FetchMangasMultiPageCSS.call(this, provider, ['/series/smaga', '/series/bmaga', '/series/others']);
return [ ...mangas1, ...mangas2 ];
```

web/src/engine/websites/SundayWebry.ts
```diff
@@ -0,0 +1,83 @@
...
+type extractChaptersResult = {
+    chapters: Chapter[],
+    nextUrl: string,
+}
+
+const chapterScript = `
+    new Promise((resolve, reject) => {
+        const listNode = document.querySelector('.js-readable-product-list');
+        const morebtn = document.querySelector('button[data-read-more-endpoint]');
+        resolve([ listNode.dataset.firstListEndpoint,morebtn.dataset.readMoreEndpoint , listNode.dataset.latestListEndpoint   ])
```
ronny1982, line 22: Spacing between variables and `,` is inconsistent

web/src/engine/websites/SundayWebry.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        //fetch first chapters
+        request = new FetchRequest(endPointlinks[0]);
+        let result = await this.extractChapters(manga, request);
+        chapters.push(...result.chapters);
+
+        //fetch MORE (paginated)
+        request = new FetchRequest(endPointlinks[1]);
+
+        for (let run = true; run;) {
```
ronny1982, line 57: A `while` or `do { ... } while` loop may be more appropriate

ronny1982 (approved):

web/src/engine/websites/SundayWebry.ts
MikeZeDev: And i though you may complain about a while loop, so i changed it to a for lmao

web/src/engine/websites/ShonenMagazine.ts
MikeZeDev: Because its dirty i didnt want to do it, knowing you will be reviewing :)
I can do it we just have to make sure the url is the pocket one at the end. Gotta try catch this to be safe.

PR #253, opened 2023-10-14
move Foolslide websites from legacy

MikeZeDev:
* FOOLSLIDE FIX  : Fix the ability to use clipboard with adult mangas.
* also blacklist twitch.tv to prevent timeouts & UI freeze on tests for some plugins


web/src/engine/websites/Mangatellers_e2e.ts
```diff
@@ -0,0 +1,25 @@
...
+        title: 'Mangatellers'
+    },
+    container: {
+        url: 'https://reader.mangatellers.gr/series/mythos/',
+        id: '/series/mythos/',
+        title: 'Mythos'
+    },
+    child: {
+        id: '/read/mythos/en/1/1/',
+        title: 'Chapter 1: Peaceful Days. (FREE)'
```
ronny1982, line 15: Is the `(FREE)` part of the title?
I would expect this to be removed.

web/src/engine/websites/decorators/FoolSlide.ts
```diff
@@ -38,7 +46,17 @@ export async function FetchMangaCSS(this: MangaScraper, provider: MangaPlugin, u
  * @param query - A CSS query to locate the element from which the manga title shall be extracted
  */
 export function MangaCSS(pattern: RegExp, query: string = queryMangaTitle) {
```
ronny1982, line 48: Have all other FoolSlide website using `import * as FoolSlide` (e.g., AnataNoMotokare, DeathTollScans, ...) been tested for regression to still work after this change?

ronny1982 (approved):

web/src/engine/websites/decorators/FoolSlide.ts
MikeZeDev: Of course boss. Both SFW and NSFW works. That say, some website now have expired certificate and couldnt be tested 

anatanomotokare : no more reader
patyscans: certificate expired
PCnet : certificate expired
silentskyscans : dead

PR #264, opened 2023-10-18
add ImageAjaxFromHTML Decorator

MikeZeDev:
and fix Assortedscans & Mexat accordingly

Mangatown needs "mangahere.com" referer for getting images, it was left untouched.

### DIFF
--- a/web/src/engine/websites/decorators/Common.ts
+++ b/web/src/engine/websites/decorators/Common.ts
@@ -457,7 +457,7 @@ export async function FetchImageAjax(this: MangaScraper, page: Page, priority: P
         const request = new FetchRequest(imageLink.href, {
             signal: signal,
             headers: {
-                Referer: page.Parameters?.Referer || page.Link.origin,
+                Referer: page.Parameters?.Referer || imageLink.origin,
             }
         });
         const response = await Fetch(request);
--- a/web/src/engine/websites/decorators/Common.ts
+++ b/web/src/engine/websites/decorators/Common.ts
@@ -4,6 +4,7 @@ import { FetchRequest, Fetch, FetchCSS, FetchWindowScript } from '../../FetchPro
 import { type MangaScraper, type DecoratableMangaScraper, type MangaPlugin, Manga, Chapter, Page } from '../../providers/MangaPlugin';
 import type { IMediaContainer } from '../../providers/MediaPlugin';
 import type { Priority } from '../../taskpool/TaskPool';
+import DeProxify from '../../transformers/ImageLinkDeProxifier';
 
 export function ThrowOnUnsupportedDecoratorContext(context: ClassDecoratorContext) {
     if (context && context.kind !== 'class') {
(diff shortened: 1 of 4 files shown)


web/src/engine/websites/decorators/Common.ts
```diff
@@ -521,6 +522,35 @@ export function ImageElement(includeRefererHeader = true, detectMimeType = false
     };
 }
 
+export async function FetchImageAjaxFromHTML(page: Page, priority: Priority, signal: AbortSignal, queryImage: string, detectMimeType = false): Promise<Blob> {
```
ronny1982, line 525: I recommend to add `this: MangaScraper` as first parameter to get type safety support

web/src/engine/websites/decorators/Common.ts
```diff
@@ -521,6 +522,35 @@ export function ImageElement(includeRefererHeader = true, detectMimeType = false
...
+    const image = await this.imageTaskPool.Add(async () => {
+        const request = new FetchRequest(page.Link.href, {
+            signal: signal
+        });
+        const realimage = (await FetchCSS<HTMLImageElement>(request, queryImage))[0].src;
+        return new Page(this, page.Parent as Chapter, new URL(DeProxify(new URL(realimage)), this.URI));
```
ronny1982, line 531: The de-proxification should be consistent with the other fetch image method.
Options:
- Remove it from this method
- Add it to all other fetch image methods
- Make it optional using a parameter for all fetch image methods

web/src/engine/websites/decorators/Common.ts
```diff
@@ -521,6 +522,35 @@ export function ImageElement(includeRefererHeader = true, detectMimeType = false
     };
 }
 
+export async function FetchImageAjaxFromHTML(page: Page, priority: Priority, signal: AbortSignal, queryImage: string, detectMimeType = false): Promise<Blob> {
+    const image = await this.imageTaskPool.Add(async () => {
+        const request = new FetchRequest(page.Link.href, {
+            signal: signal
```
ronny1982, line 528: Should a referer header be added?
```typescript
headers: {
    Referer: page.Parameters?.Referer || page.Link.origin,
}
```

ronny1982 (approved):

web/src/engine/websites/decorators/Common.ts
MikeZeDev: > Make it optional using a parameter for all fetch image methods.

Seems like the best option, with deproxify being false by default. This wont affect all Madara & MangaStream plugins since they deproxify by themselves.

web/src/engine/websites/decorators/Common.ts
MikeZeDev: I think the best option is to : 

1) Use page.Link.origin as referrer for getting page, always.
2) Use Page.parameters.Referrer OR (if it doenst exists) page.link as referer to pass in our custom fake page passed to FetchImageAjax

MikeZeDev:
Is that ok for you now?

web/src/engine/websites/decorators/Common.ts
```diff
@@ -450,10 +449,12 @@ export function PagesSinglePageJS(script: string, delay = 0) {
...
  */
-export async function FetchImageAjax(this: MangaScraper, page: Page, priority: Priority, signal: AbortSignal, detectMimeType = false): Promise<Blob> {
+export async function FetchImageAjax(this: MangaScraper, page: Page, priority: Priority, signal: AbortSignal, detectMimeType = false, deProxifyLink = false): Promise<Blob> {
     return this.imageTaskPool.Add(async () => {
-        const request = new FetchRequest(page.Link.href, {
+        const imageLink = deProxifyLink ? DeProxify(page.Link) : page.Link;
+        const request = new FetchRequest(imageLink.href, {
             signal: signal,
             headers: {
                 Referer: page.Parameters?.Referer || page.Link.origin,
```
ronny1982, line 460: Maybe use `imageLink.origin` instead of `page.Link.origin`?
Probably depends what the image proxy would set as the `Referer` header.

MikeZeDev: Hum. Seeing that we already deproxify pictures links from all madara and mangastream websites before that change,(meaning they have no referrer problem at all), and that only a few websites will use deproxify right here i'll make that change.

PR #269, opened 2023-10-19
move team1x1 from legacy

MikeZeDev:


web/src/engine/websites/Team1x1.ts
```diff
@@ -0,0 +1,47 @@
...
+    const title = anchor.querySelector('.tt').textContent.trim();
+    return {id, title };
+}
+
+@Common.MangasMultiPageCSS('/series?page={page}', 'div.bs div.bsx a', 1, 1, 0, MangaExtractor)
+@Common.PagesSinglePageCSS('.page-break img')
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+    public constructor() {
+        super('team1x1', `Team X`, 'https://team11x11.com', Tags.Language.Arabic, Tags.Media.Manhua, Tags.Media.Manhwa);
```
ronny1982, line 19: May add `Scanlator` tag if this is a scanlation group

web/src/engine/websites/Team1x1.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    public override get Icon() {
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
```
ronny1982, line 25: Can't this be replaced by the `Common.MangaCSS` class decorator?

ronny1982 (approved):

web/src/engine/websites/Team1x1.ts
MikeZeDev: This is made because they often change url without a portal page.
In Hakuneko url is a setting for this website.

Its for accommodating the future "URL" setting that should be done. Should i add a TODO to?

PR #279, opened 2023-10-21
move webace from legacy

MikeZeDev:


web/src/engine/websites/WebAce.ts
```diff
@@ -1,11 +1,12 @@
...
-import * as Common from '../decorators/Common';
-import { FetchCSS, FetchJSON, FetchRequest } from '../../FetchProvider';
+import { Chapter, DecoratableMangaScraper, Manga, Page, type MangaPlugin } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import { FetchCSS, FetchJSON, FetchRequest } from '../FetchProvider';
+
 function MangaExtractor(element: HTMLDivElement) {
     const anchor = element.querySelector<HTMLAnchorElement>('a');
-    const id = anchor.pathname.split('comics/')[0];
+    const id = anchor.pathname.match(/(\/\S+\/contents\/\d+\/)/)[1];
```
ronny1982, line 9: Reconsider the `\S+` as it will also matches `/` slashes.
Other places in this file as well...

web/src/engine/websites/WebAce_e2e.ts
```diff
@@ -0,0 +1,25 @@
...
+    },
+    entry: {
+        index: 0,
+        size: 317_431,
+        type: 'image/jpeg'
+    }
+};
+
+const fixture = new TestFixture(config);
+describe(fixture.Name, () => fixture.AssertWebsite());
```
ronny1982, line 25: My be simplified to `describe(fixture.Name, fixture.AssertWebsite);`

ronny1982 (approved):

web/src/engine/websites/WebAce_e2e.ts
MikeZeDev: hum, well

image

i did gather all changes from master tho

ronny1982: Mh, not sure how `describe` works internally, maybe it requires a closure for some access to outer variables.
Guess should be left untouched then...

PR #293, opened 2023-10-23
yugenmangas fixes

MikeZeDev:
* remove legacy Yugenmangas, which is the same website than Yugenmangas-ES in the upper level (both url were .com)

* Fix YugenMangasES & PT urls & tests

* Fix YugenmangasPT support. No madara, its custom API

* Note : YugenmangasES need novel supports, and its not madara


MikeZeDev:
> Should the bookmark importer map from `yugenmangas` to `yugenmangas-es`?

I made the update already :) BookmarkImporter is now up to date with the changed ID list.

PR #296, opened 2023-10-23
Introduce Transformer for Image De-Scrambling

ronny1982:
It is expected that there are quite a few websites using shuffled image tiles.
To avoid repetition of canvas creation and image extraction, this proposal introduces a wrapper function that handles the boilerplate.


MikeZeDev (approved):
Its a bit puzzling to me but it makes senses.
I tried using it with LineWebtoon decorator (wip) and it works fine.

I assume it could be used as long as we got either the image size (then we use new ImageData) 
or directly the blob. 

Nice job.

PR #302, opened 2023-10-28
Add ComicWalker

MikeZeDev:


web/src/engine/websites/ComicWalker.ts
```diff
@@ -0,0 +1,113 @@
...
+        return Common.GetTypedData(decrypted);
+    }
+
+    /**
+     ******************************
+     * ** COMIC-WALKER CODE BEGIN ***
+     *****************************
+     */
+
+    private generateKey(t: string): Uint8Array {
```
ronny1982, line 89: Is the previous comment inferred as JSDoc for this method?
In case it is, use a different comment block fence e.g., `/*` instead of `/**`

web/src/engine/websites/ComicWalker.ts
```diff
@@ -0,0 +1,113 @@
...
+        title: anchor.title.trim()
+    };
+}
+
+@Common.MangaCSS(/^https?:\/\/comic-walker\.com\/contents\/detail\/[^/]+\/$/, 'div#mainContent div#detailIndex div.comicIndex-box h1')
+@Common.ChaptersSinglePageCSS('div#ulreversible ul#reversible li a', ChapterExtractor)
+export default class extends DecoratableMangaScraper {
+
+    public constructor() {
+        super('comicwalker', `コミックウォーカー (ComicWalker)`, 'https://comic-walker.com', Tags.Language.Multilingual, Tags.Source.Official, Tags.Media.Manga);
```
ronny1982, line 36: Maybe add the three supported language tags instead of multilingual.

web/src/engine/websites/ComicWalker.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        //get endpoints data
+        const request = new FetchRequest(new URL(chapter.Identifier, this.URI).href);
+        const mainApp = await FetchCSS(request, 'main#app');
+        const endpoints: string | TEndpoints = mainApp[0].dataset.apiEndpointUrl ? mainApp[0].dataset.apiEndpointUrl : JSON.parse(mainApp[0].dataset.apiEndpointUrls);
+        const uri = `${(endpoints as TEndpoints).nc || (endpoints as TEndpoints).cw || endpoints}/api/v1/comicwalker/episodes/${mainApp[0].dataset.episodeId}/frames`;
+
+        const { data } = await FetchJSON<APIPages>(new FetchRequest(uri));
```
ronny1982, line 65: Will this work if a language is set which is incompatible with the chapter?
e.g., [KDCW_MF09203324020032_68](https://comic-walker.com/viewer/?tw=2&dlcl=en&cid=KDCW_MF09203324020032_68) is only available in english, what happens when last language was set to `jp` (after manga list update)?

ronny1982 (approved):

web/src/engine/websites/ComicWalker.ts
MikeZeDev: I dont think its a good idea. Let's think further : about language media tagging. 

Websites have either one langage or multiple. For now if we except special website medias have no language tag.
Now how should we implement langage tag for media? 

Website has one langage tag? > Manga directly inherit the website langage tag, in constructor. A simple constructor modification.

Website has Multilingual tag? We can then put an "unknown" tag / manually taking care of the tag.

Website has multiple langage tag, what should we do? 

Maybe i ma overthinking here.

MikeZeDev: Another idea : shouldnt we SEPARATE langage tags from the others since we put all kind of tags  in the same bag? So its easier when we have to look for langage tags only?

ronny1982: Some websites support quite a large amount of different languages.
Multilingual was introduced because of laziness, think of it as kind like a bitmask:
`type Multilingual = EN | FR | ES | JP | ZH | ...`.
If it would be feasible, i would vote for the complete removal of multilingual.

Media must not inherit default language tags from their parent, they need them explicitly set by the website plugin. In case that no language tag can be determined, another solution must be applied e.g., like you suggested with `unknown` or don't adding a language tag at all.

However, i don't think we have enough resources to implement language extraction for 800+ websites, so it is something not happening in the near future...

web/src/engine/websites/ComicWalker.ts
MikeZeDev: Right, got that fixed properly

web/src/engine/websites/ComicWalker.ts
MikeZeDev: > Media must not inherit default language tags from their parent,
If website have one language tag i dont see why? Maybe you have a better understanding than me.

> However, i don't think we have enough resources to implement language extraction for 800+ websites, so it is something not happening in the near future...

If the inheritance from the parent would be ok for you it would already be way less work.

ronny1982: I'm fine with inheriting the parents single language tag if it is beneficial. We probably may need to add more languages.

But personally I don't see a use case where a tag is really needed for every manga/chapter on a website that only provides a single language. Seems redundant to me, as the frontend could use the same logic to traverse the parents language tags on the fly.

MikeZeDev: I agree its not really "to be used" in a single language website. Its all about consistency.
Its more like that we should do the same thing for every website, and langage tagging makes sense for multilingual websites. is the word.

If the frontend can accommodate  the lack of tag, fine for me 🦐

PR #325, opened 2023-11-02
 KissComic : Rename, recaptcha support & ads blocking

MikeZeDev:
* Rename KissComic -> ReadComicOnlineLI (with relevant BookmarkConverter entries)

* Add recaptcha support for the website. https://github.com/manga-download/haruneko-proprietary/commit/1d25abf63729936716ad76455c551b051b763dbe

* Updated Blocklist so ads doesnt prevent page fetching to work.


web/src/engine/websites/ReadComicOnlineLI.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Common.MangaCSS(/^{origin}\/Comic\/[^/]+$/, 'div.barContent a.bigChar')
+@Common.MangasMultiPageCSS('/ComicList?page={page}', 'div.list-comic div.item, div.item-list div.group div.col.info', 1, 1, 0, MangInfoExtractor)
+@Common.ChaptersSinglePageCSS('div.episodeList table.listing tr td:first-of-type a, div.section ul.list li a', ChapterExtractor)
+@Common.PagesSinglePageJS('lstImages', 2500) //may trigger a captcha request
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+
+    public constructor() {
+        super('readcomiconlineli', `ReadComicOnlineLI`, 'https://readcomiconline.li', Tags.Language.English, Tags.Media.Comic, Tags.Source.Aggregator);
```
ronny1982, line 33: Consider to drop the `li` suffix since there is no other website with this name yet.

web/src/engine/websites/ReadComicOnlineLI.ts
```diff
@@ -0,0 +1,40 @@
...
+    return {
+        id: anchor.pathname,
+        title: dom.querySelector('p.title').textContent.trim()
+    };
+}
+
+@Common.MangaCSS(/^{origin}\/Comic\/[^/]+$/, 'div.barContent a.bigChar')
+@Common.MangasMultiPageCSS('/ComicList?page={page}', 'div.list-comic div.item, div.item-list div.group div.col.info', 1, 1, 0, MangInfoExtractor)
+@Common.ChaptersSinglePageCSS('div.episodeList table.listing tr td:first-of-type a, div.section ul.list li a', ChapterExtractor)
+@Common.PagesSinglePageJS('lstImages', 2500) //may trigger a captcha request
```
ronny1982, line 28: What about the parameters `readType=1` and `quality=hq`, are these no longer necessary?

web/src/engine/websites/ReadComicOnlineLI.webp
ronny1982, line 1: Is the logo still up-to-date after rebranding?

web/src/engine/websites/ReadComicOnlineLI.ts
MikeZeDev: I did that on purpose. We already got a ReadComicOnline.ru with such name.

image

And its a different website

web/src/engine/websites/ReadComicOnlineLI.ts
MikeZeDev: They are necessary when getting pages. Hence the ChapterExtractor :

image

web/src/engine/websites/ReadComicOnlineLI.webp
MikeZeDev: Yes .https://readcomiconline.li/Content/images/favicon.ico

web/src/engine/websites/ReadComicOnlineLI.ts
ronny1982: Yes the names are very similar, but the identifier `readcomiconline` is technically still different from `readcomicsonline`.
The `li` suffix may become confusing in case the website changes their top-level domain again.

web/src/engine/websites/ReadComicOnlineLI.ts
ronny1982: I see. Not a big fan of adding redundant information to the `Identifier`, but since the chapter `Identifier` isn't that much relevant in any other places, this seems to be a good solution.

ronny1982 (approved):

web/src/engine/websites/ReadComicOnlineLI.ts
MikeZeDev: I shall remove the LI then 🤷

web/src/engine/websites/ReadComicOnlineLI.ts
MikeZeDev: Yeah, i'd rather do that than overriding a method,  Its more elegant.

PR #326, opened 2023-11-02
Add Mangalife/MangaSee Websites.

MikeZeDev:
Pretty simple template used by Mangalife & MangaSee.


web/src/engine/websites/decorators/MangaLife.ts
ronny1982, line 1: I don't think there is any value to create decorators for just these two websites.
Both websites have been so similar over the last few years that it would be acceptable to use inheritance.

MikeZeDev: You are right. I am trying to do that actually without much success.

web/src/engine/websites/MangaLife.ts
```diff
@@ -0,0 +1,71 @@
+import { type Tag, Tags } from '../Tags';
+import icon from './MangaLife.webp';
+import { DecoratableMangaScraper, Manga, type MangaPlugin} from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import { FetchJSON, FetchRequest, FetchWindowScript } from '../FetchProvider';
+
+export type APIManga = {
```
ronny1982, line 7: There are many exports in this file, are these really used by other modules?

ronny1982 (approved):

web/src/engine/websites/MangaLife.ts
MikeZeDev: Its just in case, but i can remove them.

PR #333, opened 2023-11-07
Add MangAS.in

MikeZeDev:


web/src/engine/websites/MangAs.ts
```diff
@@ -0,0 +1,55 @@
...
+@Common.MangaCSS(/^{origin}\/manga\/[^/]+$/, MangaReader.queryMangaTitle)
+@Common.MangasMultiPageCSS('/filterList?page={page}&sortBy=name', 'div.media a.thumbnail', 1, 1, 500, MangaExtractor )
+@Common.ChaptersSinglePageJS(chapterScript, 500)
+@Common.PagesSinglePageCSS(MangaReader.queryPages, MangaReader.ChapterPageExtractor)
+
+export default class extends DecoratableMangaScraper {
+
+    public constructor() {
+        super('mangas', `MangAs`, 'https://mangas.in', Tags.Language.Spanish, Tags.Media.Manga, Tags.Media.Manhua, Tags.Media.Manhwa, Tags.Source.Aggregator);
+        this.Settings.throttle = new Numeric('throttle.images', R.Plugin_Settings_ThrottlingInteraction, R.Plugin_Settings_ThrottlingInteractionInfo, 4, 1, 60);
```
ronny1982, line 38: Setting description and RateLimit do not match.
Description says requests/minute but RateLimit is set to requests per 5 seconds window.

web/src/engine/websites/MangAs.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        const server1 = 's1.mangas.in/uploads/manga';
+        const server2 = 's2.mangas.in/uploads/manga';
+        const blob = await Common.FetchImageElement.call(this, page, priority, signal, true, true);
```
ronny1982, line 50: May load from both servers concurrently and use the one which returns the image (e.g., Promise.race).

ronny1982 (approved):

web/src/engine/websites/MangAs.ts
```diff
@@ -0,0 +1,55 @@
...
+        this.Settings.throttle = new Numeric('throttle.images', R.Plugin_Settings_ThrottlingInteraction, R.Plugin_Settings_ThrottlingInteractionInfo, 4, 1, 60);
+        this.Settings.throttle.ValueChanged.Subscribe((_, value: number) => this.imageTaskPool.RateLimit = new RateLimit(value, 5));
+        this.imageTaskPool.RateLimit = new RateLimit(this.Settings.throttle.value as number, 5);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
```
ronny1982, line 47: This looks very similar to code from #152
Should this PR be integrated into #152 instead?

web/src/engine/websites/MangAs.ts
MikeZeDev: I see. Are you perhaps suggesting another description?

web/src/engine/websites/MangAs.ts
MikeZeDev: It means more requests concurrently and i fear about the rate limit. The website is too unstable for that i think

web/src/engine/websites/MangAs.ts
MikeZeDev: Since this website has nothing to do with Readmangalive template i'd rather make a separate PR.

Now that you say it, maybe the FetImage from #152 should be another decorator in Common? Having to test some servers should be be uncommon.

Maybe the FetchImageElement can be replaced with FetchImageAjax? Again those two servers are driving me nuts with random 403 errors.

I think ill let it like that for now and i will test when #152  will be merged.

web/src/engine/websites/MangAs.ts
ronny1982: It would be confusing for me as user if every website uses different units for rate limit. Its probably better to just define the RateLimit in requests per minute instead of requests per 5 seconds.
`new RateLimit(..., 60);`

MikeZeDev:
Turns out the download errors were only on my end and solved with a VPN

PR #340, opened 2023-11-09
Add ElevenToon

MikeZeDev:


web/src/engine/websites/ElevenToon.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async Initialize(): Promise<void> {
+        const uri = new URL(this.URI);
+        const request = new FetchRequest(uri.href);
+        this.URI.href = await FetchWindowScript(request, `window.location.origin`, 1500);
```
ronny1982, line 30: The `uri` is changed, but not recognised in `ValidateMangaURL` as it still matches against the hardcoded URL

ronny1982 (approved):

web/src/engine/websites/ElevenToon.ts
MikeZeDev: Ah, good find. Fixing the Regexp .

MikeZeDev: I am afraid i dont get it.
>  but not recognised in ValidateMangaURL

Isnt ValidateMangaURL supposed to validate, well, manga url? Its not used to validate the website url per se.

Because its working on my end in that case

image
image
image

Ok the regexp can be improved (and i can do that) but it works. I wouldnt have submitted a PR without a working test.

Maybe i should use Common.MangaCSS ? I dont know if this website got a stable url for now. I mainly ported the code.

ronny1982: It looked to me like this is a website that changes their domain quite often, so i hinted out that the website URL may deviate with the validation pattern sometime in the future and a manual update of the pattern would still be required by a developer.
With latest commit it is now possible to use `{origin}` and `{hostname}` placeholders in the regex pattern as well.

MikeZeDev: Perfect

PR #346, opened 2023-11-21
Add ConstellarScans

MikeZeDev:


web/src/engine/websites/ConstellarScans.ts
```diff
@@ -0,0 +1,82 @@
...
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const uri = new URL(chapter.Identifier, this.URI);
+        const request = new FetchRequest(uri.href);
+        const response = await Fetch(request);
+        const data = await response.text();
+        let tsrundata = '';
+        let ts_reader: TSREADER = undefined;
+
+        //get renamed ts_reader
+        try {
```
ronny1982, line 37: Maybe it would be easier  to just get the images via JS script `div#reader_area img`

ronny1982 (approved):

web/src/engine/websites/ConstellarScans.ts
MikeZeDev: IIRC we dont have image url at all. Images src are placeholder .

PR #347, opened 2023-11-22
Add CopyManga

MikeZeDev:
Website use Url override, a setting for the image format, and a setting about the CDN.


ronny1982:
Only need to update `en_US` locale, all other will be auto-generated via Crowdin translations

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+        version: new Date().toISOString().replace(/-/g, ".").split('T').shift(),
+        platform: "1",
+    };
+
+    private readonly apiurl = 'https://api.copymanga.org';
+
+    public constructor() {
+        super('copymanga', 'CopyManga', 'https://www.copymanga.site', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Media.Manga, Tags.Language.Chinese, Tags.Source.Aggregator);
+
+        //URL
```
ronny1982, line 57: Comments are redundant, information can be inferred from code

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+        url: string
+    }[];
+    words: number[]
+}
+
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+
+    private readonly API_HEADERS = {
+        'User-Agent': '"User-Agent" to "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.5005.124 Safari/537.36 Edg/102.0.1245.44"',
```
ronny1982, line 47: Consider using `"User-Agent" to "${navigator.userAgent}"` in case the user has changed the user-agent (e.g., via HakuNeko Assistant to bypass CloudFlare)

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+        version: new Date().toISOString().replace(/-/g, ".").split('T').shift(),
```
ronny1982, line 48: This will only be created once and after two days running HakuNeko this will be quite outdated, is this by intention?
If not consider making this field a getter to evaluate and return a fresh header value.

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+        platform: "1",
+    };
+
+    private readonly apiurl = 'https://api.copymanga.org';
+
+    public constructor() {
+        super('copymanga', 'CopyManga', 'https://www.copymanga.site', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Media.Manga, Tags.Language.Chinese, Tags.Source.Aggregator);
+
+        //URL
+        this.Settings.url = new Text('urloverride', W.Plugin_Settings_UrlOverride, W.Plugin_Settings_UrlOverrideInfo, 'https://www.copymanga.site');
```
ronny1982, line 58: May simplify to use domain instead of URL, as almost all websites are using `https` and are running on port `80`

ronny1982: Consider using `this.URI.href` or `this.URI.hostname` as default value to avoid duplicated strings

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+        this.Settings.url.ValueChanged.Subscribe((_, value: string) => this.URI.href = value);
+        this.URI.href = this.Settings.url.value as string;
+
+        //image format
+        this.Settings.format = new Choice('image.format',
+            W.Plugin_Settings_ImageFormat,
+            W.Plugin_Settings_ImageFormatInfo,
+            'jpeg',
+            { key: 'jpeg', label: E.Settings_Global_DescramblingFormat_JPEG },
```
ronny1982, line 67: The used resource implies a relation to descrambling. If for some reason the text changes, it may become unrelated for this specific context.

ronny1982: Either rename the used resource, or create a new one.

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+            { key: 'webp', label: E.Settings_Global_DescramblingFormat_WEBP },
+        );
+
+        //"Use Global CDN"
+        this.Settings.useGlobalCDN = new Check(
+            'check-useGlobalCDN',
+            W.Plugin_CopyManga_Settings_GlobalCDN,
+            W.Plugin_CopyManga_Settings_GlobalCDNInfo,
+            true //true = 0, false = 1
```
ronny1982, line 76: Comment is redundant, can be directly seen in line 157

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+    };
+
+    private readonly apiurl = 'https://api.copymanga.org';
+
+    public constructor() {
+        super('copymanga', 'CopyManga', 'https://www.copymanga.site', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Media.Manga, Tags.Language.Chinese, Tags.Source.Aggregator);
+
+        //URL
+        this.Settings.url = new Text('urloverride', W.Plugin_Settings_UrlOverride, W.Plugin_Settings_UrlOverrideInfo, 'https://www.copymanga.site');
+        this.Settings.url.ValueChanged.Subscribe((_, value: string) => this.URI.href = value);
```
ronny1982, line 59: What happens to the `apiURL` when the domain is changed, should it be updated as well?

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const id = new URL(url).pathname.split('/').pop();
+        const request = this.createApiRequest(`api/v3/comic2/${id}`);
+        const data = await FetchJSON<APIResponse<APISingleComic>>(request);
+        return new Manga(this, provider, data.results.comic.path_word, data.results.comic.name.trim());
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const mangaList = [];
+        for (let page = 0, run = true; run; page++) {//its offset so page start at 0
```
ronny1982, line 98: What is meant by offset in comment?
If its important, try to give a better explanation, otherwise remove the comment.

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+            const mangas = await this.getMangasFromPage(page, provider);
+            mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+        }
+        return mangaList;
+    }
+
+    private async getMangasFromPage(page: number, provider: MangaPlugin): Promise<Manga[]> {
+        try {
+            const request = this.createApiRequest(`/api/v3/comics?ordering=-datetime_updated&limit=50&offset=${page * 50}`); //50 is the max
```
ronny1982, line 107: Comment is redundant, developers will likely always try to get as much entries within a query as possible, somi would imply that 50 is the upper limit chosen by the developer.

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+            const data = await FetchJSON<APIResponse<APIResultList<APIComic>>>(request);
+            return data.results.list.map(item => {
```
ronny1982, line 109: Can be shortened by remove redundant body
```typecript
return data.results.list.map(item => new Manga(this, provider, item.path_word, item.name.trim()));
```

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+        for (let page = 0, run = true; run; page++) {
+            const chapters = await this.getChaptersFromPage(manga, page);
+            chapters.length > 0 ? chapterList.push(...chapters) : run = false;
+        }
+        return chapterList;
+    }
+
+    private async getChaptersFromPage(manga: Manga, page: number): Promise<Chapter[]> {
+        try {
+            const request = this.createApiRequest(`/api/v3/comic/${manga.Identifier}/group/default/chapters?limit=500&offset=${page * 500}`);
```
ronny1982, line 128: Consider moving the prefix `/api/v3/` into the `createApiRequest` method.

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+            const data = await FetchJSON<APIResponse<APIResultList<APIChapter>>>(request);
+            return data.results.list.map(item => {
```
ronny1982, line 130: Body of callback method is redundant and may be removed

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const request = this.createApiRequest(`/api/v3/comic/${chapter.Parent.Identifier}/chapter2/${chapter.Identifier}?platform=3`);
+        const data = await FetchJSON<APIResponse<APIPages>>(request);
+        const imageUrls = data.results.chapter.contents.map(item => item.url);
+        const imageOrder = data.results.chapter.words;
+        const orderedPages : string[] = [];
+        imageOrder.forEach((order, index) => {
+            orderedPages[order] = imageUrls[index];
```
ronny1982, line 145: The variable `order` is of type string and not number, i would expect that it does not work correctly if this is not a number. Shouldn't it be sorted in some way to work?

web/src/engine/websites/CopyManga.ts
```diff
@@ -0,0 +1,161 @@
...
+        });
+        return orderedPages.map(page => new Page(this, chapter, new URL(page)));
+    }
+
+    private createApiRequest(pathname: string): FetchRequest {
+        const request = new FetchRequest(new URL(pathname, this.apiurl).href);
```
ronny1982, line 151: Just assign the headers directly in `FetchRequest`
```
const request = new FetchRequest(url, {
    headers: {
        'User-Agent': '...',
        'Version': '...',
        'Platform': '1',
    }
});
```

MikeZeDev:
> Only need to update `en_US` locale, all other will be auto-generated via Crowdin translations

Oh, ok, my bad i'll remember that

web/src/engine/websites/CopyManga.ts
MikeZeDev: to be frank it took this from the connector in hakuneko i dont know if a custom UA is really necessary.

ronny1982:
> > Only need to update `en_US` locale, all other will be auto-generated via Crowdin translations
> 
> Oh, ok, my bad i'll remember that

Unfortunately it was not possible to add an explicit file header comment stating which files are auto-generated.
So it can just barely be deduced from line 3~8 that `en_US` is the base and all others are auto-generated...

web/src/engine/websites/CopyManga.ts
MikeZeDev: Ressources are just filters ("*.jpeg : Jpeg image") . I think the name is just bad, yup.

Renaming them wont screw up Crowdin?

web/src/engine/websites/CopyManga.ts
MikeZeDev: Apparently apiUrl doesnt change. Idk, its like that in hakuneko.

Website use another API which is encrypted tho. Maybe this one is legacy?

web/src/engine/websites/CopyManga.ts
MikeZeDev: Offset is when you use an item count instead of a page count.
Well its not really important

web/src/engine/websites/CopyManga.ts
MikeZeDev: You could have asked anyway :D but ok.

web/src/engine/websites/CopyManga.ts
MikeZeDev: linter and editor tell me `order ` is a number  but ill double check.

MikeZeDev: Its a number there is no problem. `data.results.chapter.words` is an array of numbers after all.
There is no way to properly order image just by using their url.

web/src/engine/websites/CopyManga.ts
```diff
@@ -148,13 +136,17 @@ export default class extends DecoratableMangaScraper {
...
-        request.headers.set("webp", this.Settings.format.Value == 'jpeg' ? '0' : '1');
-        request.headers.set("region", this.Settings.useGlobalCDN.Value ? '0' : '1');
+        const request = new FetchRequest(new URL(`/api/v3/${pathname}`, this.apiurl).href,
+            {
+                headers: {
+                    Version: this.Apidate,
+                    Platform: '1',
+                    Referer: this.URI.href,
+                    Region: this.Settings.useGlobalCDN.Value ? '0' : '1',
+                    WebP: this.Settings.format.Value == 'jpeg' ? '0' : '1',
```
ronny1982, line 146: Semantically it would be more readable when using `=== 'webp' ? '1' : '0'`

ronny1982 (approved):

ronny1982:
Image is to large, test failed

MikeZeDev:
> Image is to large, test failed

Fixed by #405

PR #369, opened 2023-12-01
Missing websites from Hakuneko - Letter M (part1)

MikeZeDev:
Only noticeable website to check : Madaradex


web/src/engine/websites/MadaraDex.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const uri = new URL(chapter.Identifier, this.URI);
+        const request = new FetchRequest(uri.href);
+        const data = await FetchCSS<HTMLImageElement>(request, 'div.page-break img');
+        return data.map(element => {
+            const link = new URL(ChapterPageExtractor.call(this, element), request.url);
+            let referer = link.searchParams.get('domain') ?? this.URI.origin;
+            referer += '/';
+            return new Page(this, chapter, DeProxify(link), { Referer: referer });
```
ronny1982, line 37: Madaradex is a very specific DeProxifier that is not used in any other website.
De-Proxification should be done within this method and not via the common `ImageLinkdeProxifier`.

web/src/engine/websites/MadaraDex.ts
```diff
@@ -0,0 +1,55 @@
...
+        });
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        return this.imageTaskPool.Add(async () => {
+            const request = new FetchRequest(page.Link.href, {
+                signal: signal,
+                headers: {
+                    Referer: page.Parameters.Referer,
+                    "User-Agent": 'Mozilla/5.0 (iPod; CPU iPhone OS 14_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/87.0.4280.163 Mobile/15E148 Safari/604.1'
```
ronny1982, line 47: Is this user agent really necessary?
I would assume these images can be shown within a browser on a desktop as well.

web/src/engine/websites/MadaraDex.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+@Madara.MangaCSS(/^{origin}\/title\/[^/]+\/$/, 'meta[property="og:title"]:not([content*="Read "])')
+@Madara.MangasMultiPageAJAX()
+@Madara.ChaptersSinglePageAJAXv2()
+
+export default class extends DecoratableMangaScraper {
+
+    public constructor() {
+        super('madaradex', 'MadaraDex', 'https://madaradex.org', Tags.Media.Manhwa, Tags.Language.English, Tags.Rating.Pornographic);
```
ronny1982, line 22: Should this be tagged as `Aggregator`?
Other websites as well...

ronny1982 (approved):

web/src/engine/websites/MadaraDex.ts
MikeZeDev: For some weird reasons, if i remove this user agent, i have 100% errors rates for pictures (except proxied one works when deproxified).

I tried to exactly replicate the browser request, i even used the "copy as Fetch" feature of browser on a working request.
The request fails in Haruneko no matter what i do. This is driving me nuts. 

There is something i didnt try : see if integrated browser can get the images properly, just in case.

web/src/engine/websites/MadaraDex.ts
MikeZeDev: If website isnt "Official" or "Scanlator" should be consider it "Aggregator" by default? 🤷 

I wonder.

web/src/engine/websites/MadaraDex.ts
MikeZeDev: image

MikeZeDev: This is beyond my understanding. My investigations gives nothing. I'll stick with this user agent since it works :/

Issue #373, opened 2023-12-04
Stuck in chapter fetching until timeout (recaptcha)


MikeZeDev:
Case : Manga-link.org. Madara.

The website occasionally display a captcha when trying to fetch the chapter list. Haruneko properly open the Window, display the captcha, user fill it and its good. Except that after captcha website redirect on its homepage, not on the manga page. In fluent core the fetch chapter icon stay active and disabled until the we end up in timeout. 

During that time  we can switch to another manga and get chapters without problems since captcha was filled.

MikeZeDev:
Website changed domain and i am unable to reproduce the issue. Closing .

PR #374, opened 2023-12-04
Add Mangalink

MikeZeDev:
Little captcha to pass. https://github.com/manga-download/haruneko/issues/373

First time we are stuck until timeout, then everything works. Minor inconvenience .


web/src/engine/websites/MangaLink.ts
```diff
@@ -0,0 +1,21 @@
+import { Tags } from '../Tags';
+import icon from './MangaLink.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Madara from './decorators/WordPressMadara';
+import * as Common from './decorators/Common';
+
```
ronny1982, line 6: Avoid adding website specific anti-scraping detections into the general `AntiScrapingDetection`.
Instead, place it directly into the related file for the website:
```typescript
AddAntiScrapingDetection(async (dom) => {
    return dom.querySelector('form#lsrecaptcha-form[action*="/.lsrecap/recaptcha?"]') ? FetchRedirection.Interactive : undefined;
});
```

web/src/engine/websites/MangaLink_e2e.ts
```diff
@@ -0,0 +1,27 @@
+import { TestFixture } from '../../../test/WebsitesFixture';
+
+//THIS test may fail because of recaptcha
```
ronny1982, line 3: Consider to block comment the parts of the test that do not working like in [NeoxScan_e2e.ts](https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/NeoxScan_e2e.ts)

ronny1982 (changes requested):
Should at least move anti-scraping detection to website file

web/src/engine/websites/MangaLink.ts
MikeZeDev: Oh, ok,

MikeZeDev:
Removed captcha detection from new included js.

web/src/engine/AntiScrapingDetection.js
```diff
@@ -1,2 +1,2 @@
 /* eslint-disable -- @preserved */
```
ronny1982, line 1: Probably no longer need to submit this file as there are no changes

ronny1982 (approved):

PR #381, opened 2023-12-06
Add Mangaromance

MikeZeDev:
using the RSS feed api which is reliable, the old method isnt.


web/src/engine/websites/MangaRomance.ts
```diff
@@ -0,0 +1,64 @@
...
+    };
+}
+
+@Common.MangasSinglePageCSS('', 'div.sidebar div#Label1 div.list-label-widget-content ul li a', MangaInfoExtractor)
+@Common.PagesSinglePageCSS('article div.entry-content img')
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+
+    public constructor() {
+        super('mangaromance', `Manga Romance`, 'https://www.mangaromance.eu', Tags.Media.Manga, Tags.Media.Manhua, Tags.Media.Manhwa, Tags.Language.English, Tags.Source.Aggregator);
```
ronny1982, line 32: Cannot reach the website, is this region locked?

web/src/engine/websites/MangaRomance.ts
```diff
@@ -0,0 +1,64 @@
...
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return new RegExp(`^${this.URI.origin}/search/label/[^?]+$`).test(this.stripSearch(url));
+    }
+
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const id = new URL(url).pathname.split('/').pop();
+        const data = await FetchCSS(new FetchRequest(this.stripSearch(url)), 'div.sidebar ul li span[dir="ltr"]');
```
ronny1982, line 45: Regarding `ValidateMangaURL` the `url` cannot contain search parameters, so stripping may be redundant.

web/src/engine/websites/MangaRomance.ts
```diff
@@ -0,0 +1,64 @@
...
+        return new Manga(this, provider, id, data[0].textContent.trim());
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const request = new FetchRequest(new URL(`/feeds/posts/default/-/${manga.Identifier}?alt=json&max-results=9999`, this.URI).href);
+        const { feed } = await FetchJSON<FeedResults>(request);
+        return feed.entry.map(entry => {
+            const goodLink = entry.link.find(a => a.rel === 'alternate');
+            const title = goodLink.title.replace(manga.Title, '').replaceAll(/\[.*\]/g, '').trim();
```
ronny1982, line 54: This regex may not work as intended in case there are more than a single pair of square brackets e.g., `[001] My first Chapter [Remove Me]`

ronny1982 (approved):

web/src/engine/websites/MangaRomance.ts
MikeZeDev: Just tested, its slow but works.

web/src/engine/websites/MangaRomance.ts
MikeZeDev: nice catch

PR #382, opened 2023-12-06
Add Mangasect

MikeZeDev:


web/src/engine/websites/MangaSect.ts
```diff
@@ -0,0 +1,46 @@
...
+        const request = new FetchRequest(new URL(`/ajax/image/list/chap/${chapterid}`, this.URI).href, {
+            headers: {
+                'X-Requested-With': 'XMLHttpRequest',
+                Referer: new URL(chapter.Identifier, this.URI).href
+            }
+        });
+
+        const response = await FetchJSON<AJAXResponse>(request);
+        const dom = new DOMParser().parseFromString(response.html, 'text/html');
+        const data = [...dom.querySelectorAll<HTMLImageElement>('a.readimg img.lazy.loaded')].sort(function (a, b) {
```
ronny1982, line 37: Is sorting really necessary?

ronny1982 (approved):

web/src/engine/websites/MangaSect.ts
MikeZeDev: Yes it is. I ended up with messed up page order without, and website does the same. Perhaps a way to mess with scrapers.

PR #397, opened 2023-12-11
Add HeanCMS Template and websites

MikeZeDev:


ronny1982:
Novels shall not be handled as Manga, similar to Anime they should have their dedicated types.
However, there is no concept yet how to process Novels as the current focus is on Anime support.

MikeZeDev:
I totally agree. That said 

1) This template is interesting as the API can gives us the html of the novel chapter directly. 
2) Because the current focus is anime doesnt mean there can be a basic novel support live we had before in the meantime.

Would it be ok to disable Novel support inside and submit it? I think there is no harm letting it like .

MikeZeDev:
* Removed novel supports for now and added basic error message & handling for those.

I woudl really like not havng to pass `apiurl `on each decorator, but it would mean making a `HeanCMS `class that extends `DecoratableMangaScraper`, will adhoc functions. 

This will kinda defeat the purpose of decorators, especially when we have 2 distinct methods to get content (AjaxV1 and AjaxV2)

27136773 (commented):
There are only 3 websites which are related and would benefit from a decorator.

MikeZeDev:
> There are only 3 websites which are related and would benefit from a decorator. May consider using template or base class approach instead.

Hu? its already the case boss.

MikeZeDev:
Using proper manga filter on manga list & clipboard should be enough. Simplified it.

PR #402, opened 2023-12-13
Add ComicEarthstar

MikeZeDev:
* CoreView + GraphQL


web/src/engine/FetchProviderNodeWebkit.ts
```diff
@@ -158,10 +158,10 @@ export async function FetchCSS<T extends HTMLElement>(request: FetchRequest, que
     return [...dom.querySelectorAll(query)] as T[];
 }
 
-export async function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, variables: string): Promise<TResult> {
+export async function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, requestVariables: string = undefined): Promise<TResult> {
 
     const graphQLRequest = new FetchRequest(request.url, {
-        body: JSON.stringify({ operationName, query, variables }),
+        body: JSON.stringify({ operationName, query, variables: requestVariables ? requestVariables: {} }),
```
ronny1982, line 164: Type mismatch, `variables` is defined as string, but an empty object `{}` is used as default.

web/src/engine/websites/ComicEarthStar.ts
```diff
@@ -0,0 +1,121 @@
+import { Tags } from '../Tags';
+import icon from './ComicEarthStar.webp';
+import { DecoratableMangaScraper, Manga, type MangaPlugin } from '../providers/MangaPlugin';
+import * as CoreView from './decorators/CoreView';
+import * as Common from './decorators/Common';
+import { FetchGraphQL, FetchRequest } from '../FetchProvider';
+
+const queryOneshots = `
+            query Earthstar_Oneshot {
```
ronny1982, line 9: Indent should be 4 spaces, but is 8 spaces

web/src/engine/websites/ComicEarthStar.ts
```diff
@@ -0,0 +1,121 @@
...
+                seriesOneshot: serialGroup(groupName: "連載・読切：読切作品") {
+                    seriesSlice {
+                        seriesList {
+                            ...Earthstar_SeriesListItem_Series
+                        }
+                    }
+                }
+            }
+            fragment Earthstar_SeriesListItem_Series on Series {
```
ronny1982, line 18: Fragment is only used once, consider to embed the content directly and drop the fragment

web/src/engine/FetchProviderNodeWebkit.ts
MikeZeDev: IKR .If i use a string the request doesnt work. Teh request
Do you suggest i modify FetchGraphQL argument to pass an objet of type "unknown" so its settled ?

MikeZeDev: Changes reverted, and so its gonna need #406 merged to work :)

ronny1982 (approved):

PR #406, opened 2023-12-14
FetchGraphQL :  change "variables" argument

MikeZeDev:
All websites affected has ben tested.

### DIFF
--- a/web/src/engine/FetchProvider.ts
+++ b/web/src/engine/FetchProvider.ts
@@ -2,7 +2,7 @@ import { type PlatformInfo, Runtime, DetectPlatform, CreateUnsupportedPlatformEr
 import * as FetchProviderNodeWebkit from './FetchProviderNodeWebkit';
 import * as FetchProviderBrowser from './FetchProviderBrowser';
 import { InternalError, NotImplementedError } from './Error';
-import type { JSONElement } from '../../../node_modules/websocket-rpc/dist/types';
+import type { JSONObject } from '../../../node_modules/websocket-rpc/dist/types';
 
 export type PreloadAction = (win: typeof window, frame: typeof window) => void;
 
@@ -56,8 +56,8 @@ export function FetchCSS<T extends HTMLElement>(request: FetchRequest, query: st
     return fetchCSS<T>(request, query);
 }
 
-let fetchGraphQL: <TResult>(request: FetchRequest, operationName: string, query: string, variables: JSONElement) => Promise<TResult> = fail;
-export function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, variables: JSONElement): Promise<TResult> {
+let fetchGraphQL: <TResult>(request: FetchRequest, operationName: string, query: string, variables: JSONObject) => Promise<TResult> = fail;
+export function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, variables: JSONObject): Promise<TResult> {
     return fetchGraphQL(request, operationName, query, variables);
 }
 
--- a/web/src/engine/FetchProvider.ts
+++ b/web/src/engine/FetchProvider.ts
@@ -55,8 +55,8 @@ export function FetchCSS<T extends HTMLElement>(request: FetchRequest, query: st
     return fetchCSS<T>(request, query);
 }
 
-let fetchGraphQL: <TResult>(request: FetchRequest, operationName: string, query: string, variables : string) => Promise<TResult> = fail;
-export function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, variables : string): Promise<TResult> {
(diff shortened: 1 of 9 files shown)


web/src/engine/websites/ArgosScan.ts
```diff
@@ -55,7 +55,7 @@ export default class extends DecoratableMangaScraper {
         const vars = { id: id };
         const request = new FetchRequest(new URL('/graphql', this.URI).href);
         const operationName = 'project';
-        const data = await FetchGraphQL<ApiSingleResult<APIManga>>(request, operationName, gql, JSON.stringify(vars));
+        const data = await FetchGraphQL<ApiSingleResult<APIManga>>(request, operationName, gql, vars);
```
ronny1982, line 58: Shouldn't `vars` still be a JSON serialised string?
The implementation in `FetchGraphQL` was not updated to support this type change.

web/src/engine/FetchProviderNodeWebkit.ts
```diff
@@ -158,7 +158,7 @@ export async function FetchCSS<T extends HTMLElement>(request: FetchRequest, que
     return [...dom.querySelectorAll(query)] as T[];
 }
 
-export async function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, variables: string): Promise<TResult> {
+export async function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, variables: unknown): Promise<TResult> {
```
ronny1982, line 161: Using `unknown` is quite an anti-pattern for using type definitions.
Consider using an appropriate type or add JSDoc comment for the method parameter.

web/src/engine/FetchProvider.ts
```diff
@@ -55,8 +55,8 @@ export function FetchCSS<T extends HTMLElement>(request: FetchRequest, query: st
     return fetchCSS<T>(request, query);
 }
 
-let fetchGraphQL: <TResult>(request: FetchRequest, operationName: string, query: string, variables : string) => Promise<TResult> = fail;
-export function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, variables : string): Promise<TResult> {
+let fetchGraphQL: <TResult>(request: FetchRequest, operationName: string, query: string, variables: unknown) => Promise<TResult> = fail;
+export function FetchGraphQL<TResult>(request: FetchRequest, operationName: string, query: string, variables: unknown): Promise<TResult> {
```
ronny1982, line 59: Type `unknown` is quite an anti-pattern for using with type definitions.
Consider using an appropriate type or add JSDoc comment for the method parameter.

MikeZeDev: The thing is, by essence, the "variables" object is just an object and its definition depends on the request.

web/src/engine/websites/ArgosScan.ts
MikeZeDev: We end up getting the same result. Turns out serializing it was just because i wanted to pass it as a string . I tested all affected website and they works. And GraphQL is sensible about data types. Gotta test them again just to be sure

web/src/engine/FetchProvider.ts
ronny1982: I don't think its just an object, it must be JSON serialisable e.g., like the [JSONElement](https://github.com/manga-download/haruneko/blob/fbec2edcda3c2f1bd02f8a3cc7064a8a307b947f/web/src/engine/ipc/InterProcessCommunication.ts#L1) type definition

web/src/engine/websites/ArgosScan.ts
ronny1982: I see, so embedding a stringified JSON object within another stringified JSON object was not by intention, but a workaround.

web/src/engine/FetchProvider.ts
MikeZeDev: Its basically { x : boolean | Array | string | number } i guess, with any number of property. It could be empty too. I am wondering which type would be ok, or how to define one accuratly. Maybe JSONElement would fit ?

web/src/engine/websites/ArgosScan.ts
MikeZeDev: Its working like a charm :)

web/src/engine/FetchProvider.ts
```diff
@@ -2,6 +2,7 @@ import { type PlatformInfo, Runtime, DetectPlatform, CreateUnsupportedPlatformEr
 import * as FetchProviderNodeWebkit from './FetchProviderNodeWebkit';
 import * as FetchProviderBrowser from './FetchProviderBrowser';
 import { InternalError, NotImplementedError } from './Error';
+import type { JSONElement } from '../../../node_modules/websocket-rpc/dist/types';
```
ronny1982, line 5: Maybe it can be even more restrict and import `JSONObject` instead?

MikeZeDev: Fine

ronny1982 (approved):

PR #417, opened 2023-12-18
Add ShonenJumpRookie

MikeZeDev:


web/src/engine/websites/ShonenJumpRookie.ts
```diff
@@ -0,0 +1,57 @@
...
+        let cursor = doc.querySelector<HTMLOListElement>('ol[data-series-read-more-url]').dataset.seriesReadMoreUrl;
+
+        while (cursor) {
+            const request = new FetchRequest(new URL(cursor, this.URI).href, {
+                headers: {
+                    'X-Requested-With': 'XMLHttpRequest'
+                }
+            });
+
+            const response = await Fetch(request);
```
ronny1982, line 45: Why not using `FetchCSS`?

ronny1982 (approved):

ronny1982 (approved):

web/src/engine/websites/ShonenJumpRookie.ts
MikeZeDev: because we need the response custom header as a cursor. FetchCSS dont give access to the answer :)

PR #439, opened 2023-12-26
Add Imgur

MikeZeDev:


web/src/engine/websites/Imgur.ts
```diff
@@ -0,0 +1,45 @@
+import { Tags } from '../Tags';
+import icon from './Imgur.webp';
+import { DecoratableMangaScraper, Manga, type MangaPlugin } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import { FetchWindowScript } from '../platform/FetchProvider';
+
+const pageScript = `
```
ronny1982, line 7: Use `ScriptInjection` type
```typescript
function pageSript(): Promise<...> {
    return new Promise(resolve => {
        // ...
    });
}

File: web/src/engine/websites/Imgur.ts [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+    new Promise(resolve => {
+        const jsonData = JSON.parse(postDataJSON);
+        resolve( jsonData.media.map(image => image.url));
+    });
+`;
+
+const mangaScript = `
```
ronny1982, line 14: Use `ScriptInjection` type
```typescript
function mangaSript(): Promise<...> {
    return new Promise(resolve => {
        // ...
    });
}

ronny1982 (approved):

web/src/engine/websites/Imgur.ts
MikeZeDev: Common.PagesSinglePageJS will only accept a string.

I can do that for mangaScript tho.

PR #451, opened 2023-12-31
Add MangaFreak

MikeZeDev:
bypass Cloudflare once and it works flawlessly


ronny1982 (approved):
Approved without reviewing

PR #454, opened 2024-01-06
[UI:classic] documentation remote fetch + cache

Sheepux:
dynamic fetching of remote documentation and caches it for a week.

weakpoints:
- doesn't cache images
- use local storage (could drain quota)

### DIFF
--- a/web/src/frontend/classic/components/content-pages/Documentation.svelte
+++ b/web/src/frontend/classic/components/content-pages/Documentation.svelte
@@ -1,0 +1,94 @@
+<script lang="ts">
+    import {
+        Tile,
+        ClickableTile,
+        Breadcrumb,
+        BreadcrumbItem,
+    } from 'carbon-components-svelte';
+    import {
+        documentation,
+        type DocCategory,
+        type DocSection,
+    } from '../../stores/Documentation';
+    let selectedDocSection: DocSection;
+    let selectedDocCategory: DocCategory;
+</script>
+
+{#if selectedDocSection}
+    <Breadcrumb noTrailingSlash>
+        <BreadcrumbItem
+            href="#"
+            on:click={() => {
+                selectedDocSection = null;
+                selectedDocCategory = null;
+            }}>Documentation</BreadcrumbItem
+        >
+        <BreadcrumbItem>{selectedDocCategory?.name}</BreadcrumbItem>
+        <BreadcrumbItem isCurrentPage>{selectedDocSection?.name}</BreadcrumbItem
+        >
+    </Breadcrumb>
+    <div class="article">{@html selectedDocSection?.content.outerHTML}</div>
+{:else}
+    {#each $documentation as docCategory}
+        <div class="category">
+            <h3 class="name">{docCategory.name}</h3>
+            <div class="content">
+                {@html docCategory.description.outerHTML}
+                <div class="section">
+                    {#each docCategory.sections as docSection}
+                        <ClickableTile
+                            light
+                            on:click={() => {
+                                selectedDocCategory = docCategory;
+                                selectedDocSection = docSection;
+                            }}
+                        >
+                            {docSection.name}
+                        </ClickableTile>
+                    {/each}
+                </div>
+            </div>
+        </div>
+    {/each}
+{/if}
+
+<style>
+    .category:not(:last-child) {
+        margin-bottom: 2em;
+    }
+    .category .name {
+        text-decoration: underline;
+    }
+    .category .content {
+        margin: 1em;
+    }
+    .section {
+        display: flex;
+        flex-flow: row wrap;
+    }
+    .section > :global(*) {
+        flex: 0 1 calc(33% - 0.5em);
+        min-width: 20em;
+        margin: 0.2em;
+    }
+    .article {
+        margin-top: 1em;
+    }
+    .article :global(p) {
+        margin-bottom: 0.5em;
+    }
+    /* override carbon components */
+    .article :global(ol) {
+        list-style: auto;
+        padding-left: 2em;
+    }
+    .article :global(li) {
+        padding: 0.2em;
+    }
+    .article :global(.border) {
+        padding: 1em;
+        padding-top: 0.2em;
+        padding-bottom: 0.2em;
+        margin-bottom: 0.5em;
+    }
+</style>
--- a/web/src/frontend/classic/stores/Documentation.ts
+++ b/web/src/frontend/classic/stores/Documentation.ts
@@ -32,6 +32,16 @@ function slideElement(content:HTMLElement, split:string[]) : HTMLElement {
     const url = 'https://hakuneko.download/docs/haruneko/';
     // TODO Documentation should be cached using a service worker
     const cacheKey = `cache:${url}`;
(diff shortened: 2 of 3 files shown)


web/src/frontend/classic/stores/Documentation.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    return slide;
+}
+
+(async function () {
+    const url = 'https://hakuneko.download/docs/haruneko/';
+    // TODO Documentation should be cached using a service worker
+    const cacheKey = `cache:${url}`;
+    const cacheTimestamp = localStorage.getItem(`${cacheKey}:timestamp`);
+    if (cacheTimestamp) {
```
ronny1982, line 36: This cache strategy has some drawbacks, the preferred strategy would be online first and fallback to cache if loading online docs failed (this would also be the strategy of choice for a service worker).

web/src/frontend/classic/stores/Documentation.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const documentation = writable<DocCategory[]>([]);
+
+/**
+ * Creates a slide section by extracting HTML elements from the given content element until a specified split element is encountered.
+ * @param content - The content element from which to extract the slide section.
+ * @param split - The element name at which to stop extracting and split the slide section.
+ * @returns The created slide section element.
+ */
+function slideElement(content:HTMLElement, split:string[]) : HTMLElement {
```
ronny1982, line 21: Readability can be improved by adding space between parameter name and type declaration `content: HTMLElement`.

web/src/frontend/classic/stores/Documentation.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    const parser = new DOMParser();
+    const html = parser.parseFromString(content, 'text/html');
+    const main = html.querySelector('section');
+    const categories: DocCategory[] = [];
+    if (main) {
+        const images = main.querySelectorAll('img');
+        images.forEach((image) => {
+            const src = image.getAttribute('src');
+            if (src && !src.startsWith('http')) {
```
ronny1982, line 61: This assumes `src` to be a relative URL, but it could be a different protocol or protocol-less in which case this transformation rule will fail.

web/src/frontend/classic/stores/Documentation.ts
```diff
@@ -0,0 +1,80 @@
...
+        slide.appendChild(nextSibling.cloneNode(true));
+        nextSibling = nextSibling.nextSibling;
+    }
+    return slide;
+}
+
+(async function () {
+    const url = 'https://hakuneko.download/docs/haruneko/';
+    // TODO Documentation should be cached using a service worker
+    const cacheKey = `cache:${url}`;
```
ronny1982, line 34: Should the search parameters and fragment of the `url` be included when using it as key for caching?

web/src/frontend/classic/stores/Documentation.ts
```diff
@@ -0,0 +1,80 @@
...
+    name: string;
+    description: HTMLElement;
+    sections: DocSection[];
+}
+export type DocSection = {
+    name: string,
+    content: HTMLElement
+}
+
+export const documentation = writable<DocCategory[]>([]);
```
ronny1982, line 13: May be preferred to use a readable store, so it can only be assigned internally and not by any other components.

web/src/frontend/classic/stores/Documentation.ts
```diff
@@ -0,0 +1,80 @@
...
+    const slide = document.createElement("div");
+    let nextSibling = content.nextSibling;
+    while (nextSibling && !split.includes(nextSibling.nodeName)) {
+        slide.appendChild(nextSibling.cloneNode(true));
+        nextSibling = nextSibling.nextSibling;
+    }
+    return slide;
+}
+
+(async function () {
```
ronny1982, line 31: Parsing relies heavily on HTML structure. When the side generator is changed, the structure may not be as expected and parsing may fail or the page DOM tree may be composed dynamically.
It would be more reliable to parse the source markdown.

web/src/frontend/classic/components/content-pages/Documentation.svelte
```diff
@@ -0,0 +1,94 @@
...
+        <BreadcrumbItem isCurrentPage>{selectedDocSection?.name}</BreadcrumbItem
+        >
+    </Breadcrumb>
+    <div class="article">{@html selectedDocSection?.content.outerHTML}</div>
+{:else}
+    {#each $documentation as docCategory}
+        <div class="category">
+            <h3 class="name">{docCategory.name}</h3>
+            <div class="content">
+                {@html docCategory.description.outerHTML}
```
ronny1982, line 36: Use the script to attach the `docCategory.description` sub-tree as child nodes into the `div` container and avoid injecting HTML as string.

ronny1982 (approved):

web/src/frontend/classic/stores/Documentation.ts
Sheepux: This was intentional so that there is no request to the server each time the app is started (as the guides are displayed on the front page). Just once a week.

web/src/frontend/classic/stores/Documentation.ts
Sheepux: the structure is based on simple level2 and level3 makdown and therefore just basic h2 and h3. This wouldn't change
https://raw.githubusercontent.com/manga-download/manga-download.github.io/master/_docs/06-HaruNeko.md
But yes. the markdown could be processed client side (we need to add an extra lib then).

web/src/frontend/classic/stores/Documentation.ts
Sheepux: Oh right. It was only writable first because i did some other things like storing collapses. Forgot to switch to readable when i removed it.

web/src/frontend/classic/stores/Documentation.ts
Sheepux: not it this case but the pattern was easy to reuse. I was first  trying to cache images as well.

web/src/frontend/classic/stores/Documentation.ts
Sheepux: This is intentional as all img will be always  (at least more than probably) relative as per current documentation. Could be improved later.

web/src/frontend/classic/stores/Documentation.ts
Sheepux: should i use a svelte lib or a more generic javascript lib so that it can be used in other interfaces ?
https://www.npmjs.com/package/svelte-markdown

web/src/frontend/classic/components/content-pages/Documentation.svelte
Sheepux: this is the svelte way to embed html as @html is a special tag
https://svelte.dev/docs/special-tags#html

however this should do the trick
`<div bind:this={docCategory.description}>  `

Sheepux: trick cannot be used as the binding  requires a set function on description.
I'll stick to @html for now as rows are generated using #each

PR #455, opened 2024-01-06
Add Liliana Decorator & update websites 

MikeZeDev:
Pretty easy decorator. For websites that looks like Mangareader.to (theme "mangareader", but too generic name) or Liliana (similar look, where placeholder image for lazy-loaded images is 2 chibi-maid dancing)

Mangareader.to NOT included, as its a bit different, has multi language support and scrambled pictures. It will be added in a future PR AFTER this one.


web/src/engine/websites/decorators/Liliana.ts
```diff
@@ -0,0 +1,42 @@
+//For websites similar to MangaSect, Manga347, ManhuaPlus, ManhuaGold,  MangaReader.to,
```
ronny1982, line 1: There is no [decorator](https://www.typescriptlang.org/docs/handbook/decorators.html) or decorator factory in this file, only some constant strings are exported. Consider moving this to a more appropriate directory.

web/src/engine/websites/ManhuaPlus.ts
```diff
@@ -1,20 +1,19 @@
...
+@Common.MangaCSS(/^{origin}\/manga\/[^/]+$/, 'article header h1')
+@Common.MangasMultiPageCSS(Liliana.mangaPath, 'section div.text-center a.clamp', 1, 1, 0, Common.AnchorInfoExtractor(true))
+@Common.ChaptersSinglePageCSS(Liliana.queryChapters)
+@Common.PagesSinglePageJS(Liliana.queryPagesScript, 500)
+@Common.ImageAjax(true, true)
 export default class extends DecoratableMangaScraper {
 
     public constructor() {
-        super('manhuaplus', 'ManhuaPlus', 'https://manhuaplus.com', Tags.Media.Manhua, Tags.Media.Manhwa, Tags.Language.English);
+        super('manhuaplus', 'ManhuaPlus', 'https://manhuaplus.org', Tags.Media.Manhua, Tags.Media.Manhwa, Tags.Language.English, Tags.Source.Aggregator);
```
ronny1982, line 15: The https://manhuaplus.com Madara website is still available. The https://manhuaplus.org website seems to be a different one. Shouldn't this be a new website instead of replacing the existing one?

web/src/engine/websites/JManga.ts
```diff
@@ -0,0 +1,22 @@
...
+import * as Common from './decorators/Common';
+import * as Liliana from './decorators/Liliana';
+import { pageScript } from './RawOtaku';
+
+@Common.MangaCSS(/^{origin}\/read\/[^/]+-raw\/$/, Liliana.queryMangaTitleFromURI)
+@Common.MangasMultiPageCSS('/raw-manga/?p={page}', Liliana.queryMangas, 1, 1, 0, Common.AnchorInfoExtractor(true))
+@Common.ChaptersSinglePageCSS('div.chapters-list-ul ul li.item a.item-link', Common.AnchorInfoExtractor(true) )
+@Common.PagesSinglePageJS(pageScript, 500)
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
```
ronny1982, line 13: Is this really a Liliana based website? It looks different and does not import Liliana resources like the other websites.

web/src/engine/websites/RawOtaku.ts
```diff
@@ -0,0 +1,40 @@
...
+        );
+    })
+`;
+
+@Common.MangaCSS(/^{origin}\/read\/[^/]+-raw\/$/, Liliana.queryMangaTitleFromURI)
+@Common.MangasMultiPageCSS('/raw-manga/?p={page}', Liliana.queryMangas, 1, 1, 0, Common.AnchorInfoExtractor(true))
+@Common.ChaptersSinglePageCSS('div.chapters-list-ul ul li.item a.item-link', Common.AnchorInfoExtractor(true) )
+@Common.PagesSinglePageJS(pageScript, 500)
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
```
ronny1982, line 31: Is this really a Liliana based website? It looks different and does not import Liliana resources like the other websites.

web/src/engine/websites/RawOtaku.ts
```diff
@@ -0,0 +1,40 @@
...
+import icon from './RawOtaku.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import * as Liliana from './decorators/Liliana';
+
+export const pageScript = `
+    new Promise(resolve => {
+        const elementid = '#' + manga.lang + '-chapters li[data-number="'+ manga.name +'"]';
+        const chapterid = document.querySelector(elementid).dataset.id;
+        $.get(
```
ronny1982, line 11: Isn't the error handling function missing in this `JQuery` [get](https://api.jquery.com/jQuery.get/) request?
May also consider to use `.done()` and `.fail()` fluent methods which are equivalent to `Promise.then` and `Promise.catch`.

web/src/engine/websites/decorators/Liliana.ts
```diff
@@ -0,0 +1,42 @@
...
+                });
+            }
+            resolve(nodes.map(element => {
+                const anchorElement = element.querySelector('a.readImg');
+                return anchorElement.href ;
+            }));
+        }
+
+        const ajaxendpoint = new URL('/ajax/image/list/chap/'+ CHAPTER_ID, window.location.href);
+        fetch(ajaxendpoint, {
```
ronny1982, line 33: Consider `async/await` instead, also reject promise in case of error.

web/src/engine/websites/decorators/Liliana.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const queryPagesScript = `
+    new Promise(resolve => {
+
+        function parseResults(data) {
+            const dom = new DOMParser().parseFromString(data, 'text/html');
+            let nodes = [...dom.querySelectorAll('div.separator')];
+
+            //sort if needed
+            if (nodes[0].hasAttribute('data-index')) {
```
ronny1982, line 19: What about the case when some nodes have this attribute, and some nodes don't?

web/src/engine/websites/decorators/Liliana.ts
```diff
@@ -0,0 +1,42 @@
...
+                nodes = nodes.sort(function (a, b) {
```
ronny1982, line 20: No need to re-assign, sort mutates the original array and just returns a reference.

web/src/engine/websites/decorators/Liliana.ts
```diff
@@ -0,0 +1,42 @@
...
+    'ul#chapters-list li a.item-link',
+    'ul#myUL li.chapter a',
+].join(',');
+
+export const queryPagesScript = `
+    new Promise(resolve => {
+
+        function parseResults(data) {
+            const dom = new DOMParser().parseFromString(data, 'text/html');
+            let nodes = [...dom.querySelectorAll('div.separator')];
```
ronny1982, line 16: Use `const` instead of `let`.

web/src/engine/websites/decorators/Liliana.ts
```diff
@@ -0,0 +1,42 @@
...
+
+            //sort if needed
+            if (nodes[0].hasAttribute('data-index')) {
+                nodes = nodes.sort(function (a, b) {
+                    const za = parseInt(a.dataset.index);
+                    const zb = parseInt(b.dataset.index);
+                    return za - zb;
```
ronny1982, line 23: May return `0` in case one of the nodes doesn't have the attribute.

web/src/engine/websites/decorators/Liliana.ts
```diff
@@ -0,0 +1,42 @@
...
+                });
+            }
+            resolve(nodes.map(element => {
+                const anchorElement = element.querySelector('a.readImg');
```
ronny1982, line 27: No need for intermediate variable, return directly.

ronny1982 (changes requested):

web/src/engine/websites/decorators/Liliana.ts
MikeZeDev: template?

web/src/engine/websites/ManhuaPlus.ts
MikeZeDev: Same owner.  For a while we can see that on the .com website.
image

web/src/engine/websites/decorators/Liliana.ts
MikeZeDev: This case was never encountered. I mean, the website script itself is reordering them. In that case it would be broken.
Its all or nothing.

web/src/engine/websites/JManga.ts
MikeZeDev: As weird as it is, they share some CSS selectors that are unique like `div.anisc-detail` and others.

Anyway, like you said, liliana is not a decorator just a set of CSS and a script, importing values for it is not a big deal.

web/src/engine/websites/RawOtaku.ts
MikeZeDev: read previous answer

web/src/engine/websites/RawOtaku.ts
MikeZeDev: I merely copied the website code. fine then.

web/src/engine/websites/decorators/Liliana.ts
MikeZeDev: Handled in sorting anyway.

web/src/engine/websites/decorators/Liliana.ts
MikeZeDev: In that case i should move MH.ts from decorators to templates, and patch websites accordingly since there is no decorators

web/src/engine/websites/RawOtaku.ts
```diff
@@ -2,24 +2,21 @@ import { Tags } from '../Tags';
...
-            },
-            function(j) {
-                if (j.status == 1) {
-                    const dom = new DOMParser().parseFromString(j.html, 'text/html');
-                    resolve([...dom.querySelectorAll('div.iv-card img')].map(image => image.dataset.src));
-                } else reject();
-            },
-            "json"
-        );
+        }).fail(error => reject(error))
```
ronny1982, line 13: I suggest to use `.fail()` as last link, in the chain, to catch errors that may otherwise occur in `.done()` afterwards and won't be handled otherwise.

PR #462, opened 2024-01-12
Add Kanman

MikeZeDev:
Its the only website left that use MHXK old template.


ronny1982 (commented):
There are only 2 websites which are related and would benefit from a decorator.

MikeZeDev:
Since soudongman content now redirect to Kanman, its only Kanman 👍

PR #470, opened 2024-01-23
Add MojoPortalComic template & websites

MikeZeDev:


ronny1982 (commented):
There are only 3 websites which are related and would benefit from a decorator.

PR #477, opened 2024-01-26
Filter C1 control chars from download path

MikeZeDev:
Haruneko is partially affected by this file path sanitize problem https://github.com/manga-download/hakuneko/issues/6718.
It properly filter C0 chars but does nothing to C1 chars 

sani1
sani2

Fixes properly remove x7F-x9F range  characters . x7F being DEL its invalid too, xA0 is NBSP and its valid.

### DIFF
--- a/web/src/engine/StorageController.ts
+++ b/web/src/engine/StorageController.ts
@@ -33,8 +33,20 @@ export function SanitizeFileName(name: string): string {
         '?': '？', // https://unicode-table.com/en/FF1F/, https://unicode-table.com/en/FE56/
         '*': '＊', // https://unicode-table.com/en/FF0A/
     };
-    //https://en.wikipedia.org/wiki/C0_and_C1_control_codes
-    return name.replace(/./g, c => c.charCodeAt(0) < 32 ? '' : c.charCodeAt(0) > 0x7E && c.charCodeAt(0) < 0xA0 ? '' : lookup[c] ?? c).replace(/[\s.]+$/, '').trim() || 'untitled';
+
+    // eslint-disable-next-line no-control-regex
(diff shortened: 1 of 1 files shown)


manga-download-owner:
This solves the effect but not the cause.

Further investigation is recommend to figure out why a name contains control characters in the first place (is the title extracted with the correct encoding from the HTML file, which text extraction is used such as `innerHTML` or `textContent`, ...).

To get better support in the future i also suggest to dump the related information in chase a name contains control characters

```typescript
// eslint-disable-next-line no-control-regex
const patternControlCharsUTF8 = /[\u0000-\u001F\u007F-\u009F]/gu;

if(patternControlCharsUTF8.test(name)) {
    const sequenceCharsUTF8 = name.split('').map(c => c.charCodeAt(0).toString(16).toUpperCase().padStart(4, '0'));
    console.warn(`The filename '${name}' contains one or more invalid control characters which are going to be removed:`, sequenceCharsUTF8, '=>', name.match(patternControlCharsUTF8));
}

return name
    .replace(patternControlCharsUTF8, '')
    .replace(/./g, c => lookup[c] ?? c)
    .replace(/[\s.]+$/, '')
    .trim() || 'untitled';
 ```

MikeZeDev:
You are right but it comes from CxC API. There is not much " page encoding " to check 🤷 in a JSON result.

image

manga-download-owner:
Seems to be an encoding error in the website itself (probably caused by chinese keyboard input through the chapter upload), not a bug in HakuNeko's text extraction from the website.

manga-download-owner (approved):

ronny1982 (approved):
Seems to be an encoding error in the website itself (probably caused by chinese keyboard input through the chapter upload), not a bug in HakuNeko's text extraction from the website.

PR #479, opened 2024-01-27
Add Mangadon

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/6038


web/src/engine/websites/Mangadon.ts
```diff
@@ -0,0 +1,133 @@
...
+    public constructor() {
+        super('mangadon', 'Mangadon', 'https://mangadon.me', Tags.Media.Manhwa, Tags.Language.French, Tags.Source.Official);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return /https:\/\/mangadon\.me\/catalogs\/show\?tb=\d+$/.test(url);
```
ronny1982, line 46: May use `this.URI.origin`

web/src/engine/websites/Mangadon.ts
```diff
@@ -0,0 +1,133 @@
...
+            const request = new Request(page.Link, {
+                signal: signal,
+                headers: {
+                    Cookie: cookies.join(';'),
+                    Referer: page.Link.origin,
+                }
+            });
+
+            const response = await Fetch(request);
+            const blob = await response.blob();
```
ronny1982, line 100: Move the descrambling outside of the task-pool by returning the promise with blob. No need to block the queue any longer after response was received so next image can be downloaded as soon as possible.

ronny1982 (approved):

PR #481, opened 2024-01-28
Mangadon : Handle login

MikeZeDev:
Request Pages infos > get cookies & images url > use cookies (manually) to fetch images
If chapter is locked, we dont get cookies unless we provide an "Authorization" header in the api request.

If you are wondering, Authorization header is not used when website fetches pictures, it only use the cookies values provided by the API.


web/src/engine/websites/Mangadon.ts
```diff
@@ -67,7 +78,10 @@ export default class extends DecoratableMangaScraper {
     }
 
     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        await this.checkAuth();
```
ronny1982, line 81: Prefer local variable instead of class field `this.authToken`

web/src/engine/websites/Mangadon.ts
```diff
@@ -77,6 +91,10 @@ export default class extends DecoratableMangaScraper {
         return (data as APIItem).attributes.pages.map(page => new Page(this, chapter, new URL(page.path, this.imgCDN), { ...cookies }));
     }
 
+    async checkAuth(): Promise<void> {
```
ronny1982, line 94: May remove method that consist of a single line and is only used once

web/src/engine/websites/Mangadon.ts
```diff
@@ -67,7 +78,10 @@ export default class extends DecoratableMangaScraper {
     }
 
     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        await this.checkAuth();
+
         const request = new Request(new URL(`/api/v1/episodes/${chapter.Identifier}?params%5Bcookie_signer%5D=true&params%5Bpages%5D=true`, this.URI));
```
ronny1982, line 83: Does the `cookie_signer` parameter affect the auth mechanism?

ronny1982 (approved):

web/src/engine/websites/Mangadon.ts
MikeZeDev: No. Its the other way around.

1) You must always get cookies from the api, logged or not, free chapter or not. They are needed for the fetch request on pictures. Only for that purpose.

2) If you have no auth, you can get cookies for pictures ONLY on FREE chapters (free meaning no logging). In case of not authorized and login needed, cookies_signer object in response is empty.

3) Its the kind of api where you specify what things you wants returned.  To fetch pictures later, "cookies_signer" param is mandatory, so we get Cloudfront cookies

web/src/engine/websites/Mangadon.ts
MikeZeDev: Fine. Avoiding "called once methods" is a given ofc. I did it like that just in case , and to match other more complex auth code we may use elsewhere. Guess i was overthinking :D

PR #485, opened 2024-02-01
Update YugenmangaPT API

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/6339

### DIFF
--- a/web/src/engine/websites/YugenMangasPT.ts
+++ b/web/src/engine/websites/YugenMangasPT.ts
@@ -1,6 +1,6 @@
 import { Tags } from '../Tags';
 import icon from './YugenMangasPT.webp';
-import { Chapter, DecoratableMangaScraper, Page, MangaPlugin, Manga } from '../providers/MangaPlugin';
+import { Chapter, DecoratableMangaScraper, Page, type MangaPlugin, Manga } from '../providers/MangaPlugin';
 import * as Common from './decorators/Common';
 import { FetchJSON } from '../platform/FetchProvider';
 
@@ -12,7 +12,6 @@ type APIManga = {
     name: string,
     slug: string
 }
-
 
 type APIChapters = {
     chapters: APIChapter[]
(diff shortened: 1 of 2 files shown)


web/src/engine/websites/YugenMangasPT.ts
```diff
@@ -1,44 +1,77 @@
...
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const { chapters } = await FetchJSON<APIChapters>(new Request(new URL('/api/get_chapters_by_serie/', this.apiUrl), {
+            method: 'POST',
+            body: JSON.stringify({
+                serie_slug: manga.Identifier
+            })
+        }));
+        return chapters
+            .sort((chapterA, chapterB) => { return chapterB.name.localeCompare(chapterA.name, undefined, { numeric: true, sensitivity: 'base' }); })
```
ronny1982, line 66: Is sorting necessary? In legacy HakuNeko chapters are presented 'as is' and the UI does the sorting.

web/src/engine/websites/YugenMangasPT_e2e.ts
```diff
@@ -6,18 +6,18 @@ const config = {
         title: 'YugenMangas (PT)'
     },
     container: {
-        url: 'https://yugenmangas.org/series/para-amar-seu-inimigo/',
-        id: '/series/para-amar-seu-inimigo/',
+        url: 'https://yugenmangas.net.br/series/para-amar-seu-inimigo',
+        id: 'para-amar-seu-inimigo',
```
ronny1982, line 10: Change of ID seems to be a breaking change, will bookmarks with previous ID still work?

ronny1982 (approved):

web/src/engine/websites/YugenMangasPT_e2e.ts
MikeZeDev: They just wont anyway 🤷  . Reason is YugenMangas keeps changing everything : API, domains, paths, rebranding... IIRC before that it WASNT an api :/. Previous bookmark would be broken without us, because of yugenmangas itself.

First i was planning to use https://ikigaimangas.com.br/ as website domain, since https://yugenmangas.org/ redirect to it but its using 'https://yugenmangas.net.br/ to store pictures and for the api domain.

There is another way tho : use their rebranded name for both YugenMangas 
https://ikigaimangas.com.br/ and https://ikigaimangas.com/ (old yugenmangas.lat)

Bookmark migration wont work anyway i would have coded it otherwise.

web/src/engine/websites/YugenMangasPT.ts
MikeZeDev: Yes it is. Api order is a real mess, and its sorted on the website there is obviously some sorting underneath.
Legacy Hakuneko YugenMangas is the spanish one, we are dealing with the PT one. 

Yugenmanga two websites are a pain to follow.

PR #490, opened 2024-02-04
Add Basic Requested Websites from Hakuneko issues list

MikeZeDev:
Bunch of Madara, Mangastream, one PizzaReader, etc. Really easy websites that follow specific decorators to the letter or only Common decorators.

I am gonna take the time to review everything properly, and mark reviewed files as such. Calm down, no one is in a hurry ;)


ronny1982 (approved):
Approved without review

PR #492, opened 2024-02-05
Add NetComics

MikeZeDev:
* Removed some language from legacy list, as api revert to EN when using them (and no content available probably)


web/src/engine/websites/NetComics.ts
```diff
@@ -0,0 +1,119 @@
...
+        const mangaList : Manga[]= [];
+        for (let page = 1, run = true; run; page++) {
+            const mangas = await this.getMangasFromPage(page, provider);
+            mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+        }
+        return mangaList.distinct();
+    }
+
+    async getMangasFromPage(page: number, provider: MangaPlugin) {
+        const request = new Request(new URL(`/api/v1/title/genre?no=${page}&size=18&genre=`, this.apiUrl), {
```
ronny1982, line 82: How about fetching mangas for all languages?

web/src/engine/websites/NetComics.ts
```diff
@@ -0,0 +1,119 @@
...
+        });
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        try {
+            const request = new Request(new URL(`/api/v1/chapter/viewer/625/${chapter.Parent.Identifier}/${chapter.Identifier}?otp=`, this.apiUrl));
+            const { data } = await FetchJSON<APIResult<APIPages>>(request);
+            return data.images.map(image => new Page(this, chapter, new URL(image.image_url)));
+        } catch (e) {
+            throw Error(W.Plugin_Common_Chapter_UnavailableError);
```
ronny1982, line 104: This will catch all errors (e.g., Out of Memory, JSON Parse, ...) with the same message, is this by intention?
Would be more appropriate to check the exact conditions for the chapter being unavailable (e.g., http response 403).

web/src/engine/websites/NetComics.ts
```diff
@@ -0,0 +1,119 @@
...
+type APIPages = {
+    images: {
+        image_url: string
+    }[]
+}
+
+@Common.ImageAjax()
+
+export default class extends DecoratableMangaScraper {
+    private readonly apiUrl = 'https://beta-api.netcomics.com';
```
ronny1982, line 37: May append `/api/v1`

web/src/engine/websites/NetComics_e2e.ts
```diff
@@ -0,0 +1,49 @@
...
+const frenchfixture = new TestFixture(frenchconfig);
+describe(frenchfixture.Name, () => frenchfixture.AssertWebsite());
+
+const englishconfig = {
+    plugin: {
+        id: 'netcomics',
+        title: 'NetComics'
+    },
+    container: {
+        url: 'https://www.netcomics.com/eb/comic/some-kind-of-marriage',
```
ronny1982, line 33: Is this a typo `eb` => `en`?

ronny1982 (approved):

web/src/engine/websites/NetComics.ts
MikeZeDev: Hum, why not

web/src/engine/websites/NetComics_e2e.ts
MikeZeDev: Oops

PR #495, opened 2024-02-07
Remove Assorted Scans

ObserverOfTime:
See manga-download/hakuneko#6764


MikeZeDev:
We are discussing the removal. Conditions seems ok.

PR #502, opened 2024-02-14
Remove ending hyphen in manga path

MikeZeDev:
* Windows allows ending path with ~ 
image

* For some reason, the Javascript API doesnt !
image

* Removing the ending "~" solves the problem 🤷 
image

Was reported on Discord.


Sheepux:
Should add testcase in tests with sheepscan Media id 30 and chapter 2
https://github.com/manga-download/sample-websites/blob/main/sheep-scanlations/index.json#L31
https://github.com/manga-download/sample-websites/blob/main/sheep-scanlations/030.json
`    { "id": "/030.json", "title": "Real 28 - €$£&éê@§% / '\"\\!?,;.~*`^°¤¶ ~" },`
`        "title": "Chapter 2 -  €$£&éê@§% / '\"\\!?,;.~*`^°¤¶"`

web/src/engine/StorageController.ts
```diff
@@ -46,6 +46,7 @@ export function SanitizeFileName(name: string): string {
         .replace(patternControlCharsUTF8, '')
         .replace(/./g, c => lookup[c] ?? c)
         .replace(/[\s.]+$/, '')
+        .replace(/~$/, '')//While Windows allows it, it seems NodeJS somehow dont like paths with hyphen at the end
```
ronny1982, line 49: Comment is incorrect, NodeJS can handle trailing tilde, the Browser's [File System API](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle) cannot.
Use a more appropriate replacement (e.g., [～](https://unicode-explorer.com/c/FF5E))
➔ Probably best to replace all tilde characters with a unicode variant to keep consistence.

ronny1982 (changes requested):

web/src/engine/StorageController.ts
MikeZeDev: Aaaaah thats the reason, thanks !

ronny1982 (approved):

PR #505, opened 2024-02-19
Add Mangaship

MikeZeDev:


web/src/engine/websites/MangaShip.ts
```diff
@@ -0,0 +1,61 @@
+import { Tags } from '../Tags';
+import icon from './MangaShip.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+
+const pageScript = `
+    new Promise(async (resolve) => {
```
ronny1982, line 7: When using an `async` callback, the automatic rejection will not work in case of an error. Make sure to catch errors and reject the promise gracefully.
```typescript
new Promise(async (resolve, reject) => {
    try {
        throw new Error('...');
    } catch(error) {
        reject(error);
    }
});
```

ronny1982 (approved):

web/src/engine/websites/MangaShip.ts
MikeZeDev: Ok

PR #522, opened 2024-03-02
HentaiRead : minor fixes

MikeZeDev:
* try catch in promise
* deproxify images
* fix chapter link in test

### DIFF
--- a/web/src/engine/websites/HentaiRead.ts
+++ b/web/src/engine/websites/HentaiRead.ts
@@ -3,26 +3,31 @@ import icon from './HentaiRead.webp';
 import { DecoratableMangaScraper } from '../providers/MangaPlugin';
 import * as Common from './decorators/Common';
 
-const pageScript = `new Promise(resolve => {
-    const pagelist = chapter_preloaded_images;
-    resolve(pagelist.map(link => {
-        let uri = new URL(link);
-        uri.searchParams.set('quality', '100');
-        uri.searchParams.delete('w');
-        return uri.href;
-    }));
-});
+const pageScript = `
+    new Promise((resolve, reject) => {
+        try {
+            const pagelist = (window.chapterImages ?? window.chapter_preloaded_images);
+            resolve(pagelist.map(image => {
+                const uri = new URL(image.src);
+                uri.searchParams.set('quality', '100');
+                uri.searchParams.delete('w');
+                return uri.href;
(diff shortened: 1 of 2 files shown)


web/src/engine/websites/HentaiRead.ts
```diff
@@ -3,26 +3,31 @@ import icon from './HentaiRead.webp';
...
-    resolve(pagelist.map(link => {
-        let uri = new URL(link);
-        uri.searchParams.set('quality', '100');
-        uri.searchParams.delete('w');
-        return uri.href;
-    }));
-});
+const pageScript = `
+    new Promise((resolve, reject) => {
+        try {
```
ronny1982, line 8: No need for `try ... catch` in non-async sequential callback.
Errors are captured directly by the promise and will reject it.

Make sure to understand the capture scope for errors within a promise and use `try ... catch` only when appropriate or necessary
```javascript
// Bad (redundant)
new Promise((resolve, reject) => {
    // Errors within promise scope can be manually captured => manually rejected
    try {
        throw new Error();
        resolve();
    } catch(error) {
        reject(error);
    }
});

// Good
new Promise(resolve => {
    // Errors within promise scope are captured automatically => automatically rejected
    throw new Error();
    resolve();
});

// Bad
new Promise(resolve => {
    // Errors outside of promise scope are not captured automatically => never rejected
    setTimeout(() => {
        throw Error();
        resolve();
    }, 0);
});

// Good
new Promise((resolve, reject) => {
    // Errors outside of promise scope must be captured manually => manually rejected
    setTimeout(() => {
        try {
            throw Error();
            resolve();
        } catch(error) {
            reject(error);
        }
    }, 0);
});

// Bad
new Promise(async (resolve) => {
    // Errors outside of promise scope are not captured automatically => never rejected
    throw Error();
    resolve();
});

// Good
new Promise(async (resolve, reject) => {
    // Errors outside of promise scope must be captured manually => manually rejected
    try {
        throw Error();
        resolve();
    } catch(error) {
        reject(error);
    }
});
```

ronny1982 (approved):

web/src/engine/websites/HentaiRead.ts
MikeZeDev: I see, its a scope problem. I think i get it now. Sorry about that.

PR #531, opened 2024-03-08
[tumangaonline] add website support

ronny1982:
See: https://github.com/manga-download/haruneko-proprietary/commit/edd21fbc145fd119e6d0cba3b27ba7a98ba5f150


app/nw/scripts/build-app.mjs
```diff
@@ -23,6 +23,10 @@ const manifest = {
         'http://localhost/*',
         'https://haruneko.pages.dev/*'
     ],
+    // See: https://peter.sh/experiments/chromium-command-line-switches/
+    'chromium-args': [
+        '--ignore-certificate-errors',
```
MikeZeDev, line 28: Arent there security implications ? Should we leave it like that ?

MikeZeDev:
If scanlator or language ares missing (since ? syntax was used) wont the title end like " [undefined] (undefined)'?

web/src/engine/websites/TuMangaOnline.d.ts
```diff
@@ -0,0 +1,7 @@
+import type { Page } from '../providers/MangaPlugin';
```
MikeZeDev, line 1: Just for knowledge purpose, whats the point of this file? Should i do the same for https://github.com/manga-download/haruneko/pull/528 ?

MikeZeDev (approved):
I am concerned about the security implications of the command line switch but LGTM.

web/src/engine/websites/TuMangaOnline.d.ts
ronny1982: After compilation and obfuscation there is only a JavaScript file. This file exposes the type definitions which are missing in the JavaScript file for better integration in a TypeScript project.

This file is automatically created when executing `npm run deploy:prod` (assuming the relative paths of haruneko and haruneko-proprietary are in the same parent directory).

It is strongly recommend to add this file.

app/nw/scripts/build-app.mjs
ronny1982: It is not safe, but required since TMO has an invalid certificate which must be manually allowed in browser. Since users (hopefully) don't use HaruNeko for banking or other important account logins, it can be considered as very low risk.

ronny1982:
> If scanlator or language are missing (since ? syntax was used) wont the title end like " Sheep Manga [undefined] (undefined)'?

Addressed in https://github.com/manga-download/haruneko-proprietary/commit/01b3fcdef6f9623934c3dcf65fff7df053255cb7

app/nw/scripts/build-app.mjs
ronny1982: In the meantime TMO fixed their untrusted certificate.
Removed permission for untrusted certificates in c2df72d3

web/src/engine/websites/TuMangaOnline.d.ts
MikeZeDev: Gotta do that for mangadig then, thanks

PR #541, opened 2024-03-18
Add Qimeitj

MikeZeDev:
Standard MH website BUT page list is paginated so a separate PR

https://github.com/manga-download/hakuneko/issues/6135


web/src/engine/websites/Qimeitj.ts
```diff
@@ -0,0 +1,41 @@
...
+        const pagesList: Page[] = [];
+        for (let page = 1, run = true; run; page++) {
+            const pages = await this.getPagesFromChapterPage(page, chapter);
+            pages.length > 0 ? pagesList.push(...pages) : run = false;
+        }
+        return pagesList;
+    }
+    async getPagesFromChapterPage(page: number, chapter: Chapter): Promise<Page[]> {
+        const url = new URL(chapter.Identifier, this.URI);
+        url.search = new URLSearchParams({
```
ronny1982, line 32: Since this is just a single parameter `url.searchParam.set()` seems to be more appropriate

ronny1982 (approved):

PR #542, opened 2024-03-18
Add Laimanhua8

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/6157


web/src/engine/websites/Laimanhua8.ts
```diff
@@ -0,0 +1,38 @@
...
+        super('laimanhua8', 'Laimanhua8', 'https://www.laimanhua8.com', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Language.Chinese, Tags.Source.Aggregator);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const mangalist : Manga[] = [];
+        const paths = 'abcdefghijklmnopqrstuvwxyz'.split('');
```
ronny1982, line 31: Suggested naming `letters`, `alphabet` or `pathSegments`.
Alternatively map to path, so the variable name expresses their correct meaning `const paths = 'abcdef'.map(letter => '/kanmanhua/${letter}.html'`

ronny1982 (approved):

PR #543, opened 2024-03-19
Add Delitoon websites (Delitoon/X/DE/Bomtoon/Bontoon/LezhinES)

MikeZeDev:
Most files are empty shells that just extends Delitoon :D


web/src/engine/websites/LezhinES.ts
```diff
@@ -0,0 +1,11 @@
+import { Tags } from '../Tags';
+import icon from './LezhinES.webp';
+import Delitoon from './Delitoon';
+export default class extends Delitoon {
```
ronny1982, line 4: LezhinES is based on Delitoon and not affiliated with LezhinEN, LezhinKO and LezhinJA?

web/src/engine/websites/Delitoon.ts
```diff
@@ -0,0 +1,174 @@
...
+        expiredAt: number
+    },
+    refreshToken: {
+        token: string
+    }
+}
+
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+    protected BalconyID: string = undefined;
```
ronny1982, line 57: Make fields private readonly and assign in constructor

web/src/engine/websites/Delitoon.ts
```diff
@@ -0,0 +1,174 @@
...
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return new RegExp(`^${this.URI.origin}/detail/[^/]+$`).test(url);
+    }
+
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const mangaid = new URL(url).href.match(/\/detail\/([^/]+)/)[1];
+        const req = new URL(`/api/balcony-api-v2/contents/${mangaid}`, this.URI);
```
ronny1982, line 77: Variable name is misleading, i would expect a Request but it is an URL

web/src/engine/websites/Delitoon.ts
```diff
@@ -0,0 +1,174 @@
...
+                const url = new URL('/api/balcony/auth/refresh', this.URI);
+                const body = JSON.stringify({
+                    accessToken: this.Token.accessToken.token,
+                    refreshToken: this.Token.refreshToken.token
+                });
+                const { result } = await FetchJSON<APIRefreshToken>(this.CreatePostRequest(url, body));
+                this.Token = result;
+            }
+        } catch (error) {
+            this.Token = undefined;
```
ronny1982, line 171: Shouldn't the exception be forwarded to be handled by the caller?
All errors such as network error, out of memory, ... are just silently ignored

web/src/engine/websites/Delitoon.ts
```diff
@@ -0,0 +1,174 @@
...
+        };
+        if (this.Token) headers['authorization'] = ' Bearer ' + this.Token.accessToken.token;
+        return new Request(url, {
+            method: 'POST',
+            headers: headers,
+            body: body,
+        });
+    }
+
+    protected async GetToken() {
```
ronny1982, line 149: Method is called `GetToken` but is void.
Use a more appropriate name for the method.

web/src/engine/websites/Delitoon.ts
```diff
@@ -0,0 +1,174 @@
...
+            //if token is undefined => try to fetch one
+            if (!this.Token) {
+                const url = new URL('/api/auth/session', this.URI);
+                const { user } = await FetchJSON<APIUser>(this.CreateRequest(url));
+                this.Token = user;
+                return;
+            }
+
+            //if token is defined, check for expiration  and refresh if needed
+            if (this.Token && Date.now() > this.Token.accessToken.expiredAt) { //24h expiration
```
ronny1982, line 161: The check for `this.Token` is redundant as it is guarded by the previous condition which checks for `!this.Token`.

web/src/engine/websites/Delitoon.ts
```diff
@@ -0,0 +1,174 @@
...
+            contentsThumbnailType: 'MAIN'
+        }).toString();
+        const { data } = await FetchJSON<APIResult<APIManga[]>>(this.CreateRequest(url));
+        return data.map(element => new Manga(this, provider, element.alias, element.title.trim()));
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        await this.GetToken();
+        const url = new URL(`/api/balcony-api-v2/contents/${manga.Identifier}`, this.URI);
+        url.searchParams.set('isNotLoginAdult', 'true');
```
ronny1982, line 98: Does this exclude all adult content?
How is adult content accessible.

web/src/engine/websites/Delitoon.ts
```diff
@@ -0,0 +1,174 @@
...
+            let title = element.title.trim();
+            title += element.subTitle ? ' : ' + element.subTitle.trim() : '';
+            return new Chapter(this, manga, element.alias, title);
+        });
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        await this.GetToken();
+        const url = new URL(`/api/balcony-api-v2/contents/${chapter.Parent.Identifier}/${chapter.Identifier}`, this.URI);
+        url.searchParams.set('isNotLoginAdult', 'true');
```
ronny1982, line 110: Is this parameter required when fetching pages?

web/src/engine/websites/Delitoon.ts
ronny1982, line 161: Is this check timezone safe, or is the expiration from the server's timezone?

ronny1982 (approved):

web/src/engine/websites/LezhinES.ts
MikeZeDev: Exactly. Weird isnt it?

web/src/engine/websites/Delitoon.ts
MikeZeDev: No. From what i see, its a variable to get adult content without being logged.

web/src/engine/websites/Delitoon.ts
MikeZeDev: image

web/src/engine/websites/Delitoon.ts
MikeZeDev: expiredAt is UTC timestamp. Date.now is the same iirc.

web/src/engine/websites/Delitoon.ts
MikeZeDev: It shows that /viewer/ is now a valid endpoint for, perhaps all websites. Gotta correct that (old still working)

PR #544, opened 2024-03-19
Add Manga Hot

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/6255

Scrambling and very bad API.


web/src/engine/websites/MangaHot.ts
```diff
@@ -0,0 +1,193 @@
...
+    public constructor() {
+        super('mangahot', 'マンガほっと (Manga Hot)', 'https://web.mangahot.jp', Tags.Media.Manga, Tags.Language.Japanese, Tags.Source.Official);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return new RegExp(`^${this.URI.origin}/works/detail.php\\?work_code=[^/]+`).test(url);
```
ronny1982, line 38: Regex will not work for valid url such as `https://web.mangahot.jp/works/detail.php?page=7&work_code=solo-leveling`.
It is not always appropriate to use a regular expression, there are other methods as well e.g.,
```typescript
const uri = new URL(url);
return uri.origin === this.URI.origin && uri.pathname === '/works/detail.php' && uri.searchParams.has('work_code');
```

web/src/engine/websites/MangaHot.ts
```diff
@@ -0,0 +1,193 @@
...
+            mode: 'ajax_stories',
+            _: Date.now().toString()
+        }).toString();
+
+        const { result } = await FetchJSON<APIResult<string>>(new Request(url, {
+            headers: {
+                'X-Requested-With': 'XMLHttpRequest',
+            }
+        }));
+        const dom = new DOMParser().parseFromString(result, 'text/html');
```
ronny1982, line 59: Consider using a more lightweight [approach](https://grrr.tech/posts/create-dom-node-from-html-string/#performance) for creating a sub-tree

web/src/engine/websites/MangaHot.ts
```diff
@@ -0,0 +1,193 @@
...
+                method: 'POST',
+                body: `story_id=${chapter.Identifier}&page_no=${page}`,
+                headers: {
+                    'X-Requested-With': 'XMLHttpRequest',
+                    'Content-type': 'application/x-www-form-urlencoded; charset=UTF-8'
+                }
+            });
+
+            const { result } = await FetchJSON<APIResult<string[] | PageResult>>(request); //result can be an array or a json object list
+            if (!result || Array.isArray(result) && result.length == 0) {
```
ronny1982, line 81: Prefer the type-safe comparison operator `===`, `'0' == 0` vs. `'0' === 0`

ronny1982: Check can be simplified
```typescript
if(!result || Object.keys(result).length === 0) {}
```

ronny1982 (approved):

web/src/engine/websites/MangaHot.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const [data] = await FetchCSS(new Request(url), 'div.work_base div.work_name');
+        return new Manga(this, provider, new URL(url).searchParams.get('work_code'), data.textContent.trim());
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const url = new URL(`/works/detail.php`, this.URI);
+        url.search = new URLSearchParams({
```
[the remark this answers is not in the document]
MikeZeDev: It means setting them one after another.  You cant UrlsearchParam.set more than one parameter at a time .
I think its more elegant that way.

web/src/engine/websites/MangaHot.ts
MikeZeDev: Hum. Tried it but i fail to see how its able to get nodes from a string.

```
        const dom = document.createDocumentFragment();
        dom.append('result')
```

dont add the proper DOM nodes.

ronny1982: Should be quite simple such as
```typescript
const container = document.createElement('template');
container.innerHTML = '<ul><li>a</li><li>b</li></ul>';
const elements = [ ...container.content.querySelectorAll('li') ];
```

MikeZeDev: Done

ronny1982: My advise was bad.
The `DOMParser` performs the same or even faster than `innerHTML` and is also more secure to cross scripting attacks.

MikeZeDev: Its the intention that matters boss.

PR #548, opened 2024-03-24
Add VortexScans

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/6499


web/src/engine/websites/VortexScans.ts
```diff
@@ -0,0 +1,105 @@
...
+        }[]
+    }
+}
+
+const mangaIdScript = `
+    new Promise((resolve, reject) => {
+       const myregexp = /"postId"\\s*:\\s*(\\d+)/i;
+        __next_f.forEach(element => {
+            const el = element[1];
+            if (el) {
```
ronny1982, line 48: Use short circuited conditions instead of nested `if`
```javascript
if (condition1 && contidion2 && condition3 && ...)
```

web/src/engine/websites/VortexScans.ts
```diff
@@ -0,0 +1,105 @@
...
+                if((myregexp.test(el)))  {
+                   const id = parseInt(el.match(myregexp)[1]);
+                   const title = document.querySelector('meta[property="og:title"]').content.trim();
+                   const slug = window.location.pathname.split('/').pop();
+                   resolve({id: id, title: title, slug: slug});
```
ronny1982, line 53: A promise should only be resolved/rejected once. Consider using a for loop instead of `.forEach` and `return resolve(...)`.

ronny1982: Or utilize `__next_f.find(el => condition1 && condition2 %% ...)`

web/src/engine/websites/VortexScans.ts
```diff
@@ -0,0 +1,105 @@
...
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const data = await FetchWindowScript<MangaID>(new Request(url), mangaIdScript, 500);
+        return new Manga(this, provider, JSON.stringify({ id: data.id, slug: data.slug }), data.title);
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const endpoint = new URL('query?perPage=9999', this.apiURL);
+        const { posts } = await FetchJSON<APIMangas>(new Request(endpoint));
+        return posts.map(manga => new Manga(this, provider, JSON.stringify({
+            id: manga.id.toString(),
```
ronny1982, line 87: In `FetchManga(` the manga id is a number (extracted with `parseInt`), here it is converted to a string. The type of the manga id should be consistent.

ronny1982:
I'm confused why VortexScans is not already part of HaruNeko, it exist in HakuNeko for 5 years.
Maybe there are more websites that are present in HakuNeko, but missing in HaruNeko...

ronny1982 (approved):

MikeZeDev:
> I'm confused why VortexScans is not already part of HaruNeko, it exist in HakuNeko for 5 years. Maybe there are more websites that are present in HakuNeko, but missing in HaruNeko...

VortexScans was dead imho. The original request was named VexMangas but Vexmangas redirect to a now alive VortexScans 🤷

web/src/engine/websites/VortexScans.ts
```diff
@@ -43,18 +43,12 @@ type APIPages = {
...
-                   const id = parseInt(el.match(myregexp)[1]);
-                   const title = document.querySelector('meta[property="og:title"]').content.trim();
-                   const slug = window.location.pathname.split('/').pop();
-                   resolve({id: id, title: title, slug: slug});
-                }
-            }
-        });
-        reject();
+       const element = __next_f.find(el => el[1] && el[1].match(myregexp));
+       if (!element) reject("Regex didnt match :/");
```
ronny1982, line 47: Code after the `if` is still executed, even after the promise was already rejected.

MikeZeDev: I guess a throw would be ok then?

ronny1982: This would work

PR #567, opened 2024-04-13
DoujinDesu : fix manga list and pages

MikeZeDev:
use fetch in page context to avoid 403 error


ronny1982:
May just add the missing headers to avoid 403
```
'X-Requested-With': 'XMLHttpRequest'
'Referer': 'chapter url'
```

MikeZeDev:
> May just add the missing headers to avoid 403
> 
> ```
> 'X-Requested-With': 'XMLHttpRequest'
> 'Referer': 'chapter url'
> ```

Its what i tried, i still got mysterious 403. With both of them and even user agent that matched my browser.
Otherwise i wouldnt have tried this solution. Maybe i did it wrong.

PR #579, opened 2024-04-20
Add Anchira

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/5483


web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+        for (let page = 1, run = true; run; page++) {
+            await reducer;
+            reducer = new Promise(resolve => setTimeout(resolve, throttle));
+            const mangas = await this.getMangasFromPage(page, provider);
+            mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+        }
+        return mangaList;
+    }
+
+    private async getMangasFromPage(page: number, provider: MangaPlugin): Promise<Manga[]> {
```
ronny1982, line 67: Use PascalCase

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+    title: string
+    hash?: string,
+    names?: string[]
+}
+
+@Common.ChaptersUniqueFromManga()
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+
+    private readonly apiUrl = `${this.URI}api/v1/`;
```
ronny1982, line 23: Is there a slash missing after `URI`?

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+    public override get Icon() {
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return new RegExp(`^${this.URI.origin}/g/[\\d]+/[^/]+$`).test(url);
+    }
+
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const matches = url.match(/\/g\/(\d+)\/([^/]+)/);
```
ronny1982, line 39: Code duplication, use the same RegExp for validating and fetching manga

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+        const id = matches[1];
+        const key = matches[2];
```
ronny1982, line 41: May just get id + key of the url path in a single capture group, no need for two separate capture groups

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+        const item = await FetchJSON<APIManga>(new Request(new URL(`${this.apiUrl}library/${id}/${key}`), {
+            headers: {
+                'X-Requested-With': 'XMLHttpRequest',
+                'Referer': `${this.URI}?`
+            }
+        }));
+        return new Manga(this, provider, JSON.stringify({
```
ronny1982, line 48: Manga identifier can be simplified to string `'id/key'` instead of `{ "id": 123, "key": "..." }` as they are joined together anyway.

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+            id: item.id.toString(),
+            key: item.key
+        }), item.title.trim());
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const mangaList = [];
+        const throttle = 500;
+        let reducer = Promise.resolve();
```
ronny1982, line 57: Use a rate limited task pool such as in Piccoma or old MangaDex

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+            const uri = new URL(`${this.apiUrl}library?page=${page}`);
+            if (page == 1) uri.searchParams.delete('page');
+            const { entries } = await FetchJSON<APIMangas>(new Request(uri, {
+                headers: {
+                    'X-Requested-With': 'XMLHttpRequest',
+                    'Referer': `${this.URI}?`
+                }
+            }));
+            return entries.map(item => new Manga(this, provider, JSON.stringify({
+                id: item.id.toString(),
```
ronny1982, line 78: Manga identifier can be simplified as described above

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+                key: item.key
+            }), item.title.trim()));
+        }
+        catch (error) {
+            return [];
```
ronny1982, line 83: Why are errors suppressed?
Wouldn't it be better to let them bubble up to the caller?

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const mangaid: APIManga = JSON.parse(chapter.Identifier);
+        const data = await FetchJSON<APIManga>(new Request(new URL(`${this.apiUrl}library/${mangaid.id}/${mangaid.key}/data`), {
```
ronny1982, line 90: Wrapping the full composed URL string into an intermediate URL object is redundant as `Request` accepts both types

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,102 @@
...
+            headers: {
+                'X-Requested-With': 'XMLHttpRequest',
+                'Referer': `${this.URI}/g/${mangaid.id}/${mangaid.key}`
+            }
+        }));
+        return data.names.map(image => {
+            const pageUrl = new URL(`/${data.id}/${data.key}/${data.hash}/a/${image}`, this.imageCDN);// /a/ ensure best quality. /b/ is lower
```
ronny1982, line 97: Remove comment that only contains ancillary information

ronny1982 (approved):

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,95 @@
...
+    title: string
+    hash?: string,
+    names?: string[]
+}
+
+@Common.ChaptersUniqueFromManga()
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+
+    private readonly mangasTaskPool = new TaskPool(4, new RateLimit(8, 1));
```
ronny1982, line 25: The provided rate limit does not match the previous value, should be `2` requests per `1` second instead of `8` requests per `1` second.

web/src/engine/websites/Anchira.ts
```diff
@@ -0,0 +1,95 @@
...
+        }));
+        return new Manga(this, provider, `${item.id}/${item.key}`, item.title.trim());
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const cancellator = new AbortController();
+        try {
+            const mangaList = [];
+            for (let page = 1, run = true; run; page++) {
+                const mangas = await this.GetMangasFromPage(page, provider, cancellator.signal);
```
ronny1982, line 58: Using a `taskpool` is redundant when only adding a new task after the previous task is completed.
The intended use is to push all tasks at once into the `taskpool` and then await all the promises to be fulfilled.

See: https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/Piccoma.ts#L67-L71

web/src/engine/websites/Anchira.ts
MikeZeDev: OK.

web/src/engine/websites/Anchira.ts
MikeZeDev: I get it now, sorry

PR #603, opened 2024-05-06
Add ComicFans

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/6876


web/src/engine/websites/ComicFans.ts
```diff
@@ -0,0 +1,94 @@
...
+
+type APIManga = {
+    id: number,
+    title: string
+}
+
+type APIChapter = {
+    id: number,
+    title: string,
+    chapterOrder: number
```
ronny1982, line 25: Can this order number really be used in the title or is it just for internal sorting?

ronny1982 (approved):

web/src/engine/websites/ComicFans.ts
MikeZeDev: Without it there is no chapter number in titles, in other words its a mess.

PR #610, opened 2024-05-14
Add NamiComi

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7078


web/src/engine/websites/NamiComi.ts
```diff
@@ -0,0 +1,111 @@
...
+    high: {
+        filename: string
+    }[]
+}
+
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+
+    private readonly apiUrl = 'https://api.namicomi.com';
+    private readonly mangaRegexp = new RegExp(`^${this.URI.origin}/en/title/(\[^/]+)\/[^/]+$`);
```
ronny1982, line 45: Website is tagged as multilingual, but regex only matches oath `/en/` is this by intention?

web/src/engine/websites/NamiComi.ts
```diff
@@ -0,0 +1,111 @@
...
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const mangaList = [];
+        for (let page = 0, run = true; run; page++) {
+            const mangas = await this.GetMangasFromPage(page, provider);
+            mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+        }
+        return mangaList;
+    }
+
+    private async GetMangasFromPage(page: number, provider: MangaPlugin): Promise<Manga[]> {
```
ronny1982, line 75: No need to get single pages for now. Website has `831` titles, so may just use a limit of `9999` to get all at once.

web/src/engine/websites/NamiComi.ts
```diff
@@ -0,0 +1,111 @@
...
+        const chapterList = [];
+        for (let page = 0, run = true; run; page++) {
+            const chapters = await this.GetChaptersFromPage(manga, page);
+            chapters.length > 0 ? chapterList.push(...chapters) : run = false;
+        }
+        return chapterList;
+    }
+
+    private async GetChaptersFromPage(manga: Manga, page: number): Promise<Chapter[]> {
+        const request = new Request(new URL(`/chapter?titleId=${manga.Identifier}&limit=100&offset=${page * 100}`, this.apiUrl));
```
ronny1982, line 94: Can limit be increased to get all chapters at once instead of paginated?

web/src/engine/websites/NamiComi.ts
```diff
@@ -0,0 +1,111 @@
...
+        const { data } = await FetchJSON<APIResult<APIChapter[]>>(request);
+        return data.map(item => {
+            let title = item.attributes.name ? `${item.attributes.chapter} ${item.attributes.name}` : `Volume ${item.attributes.volume} Episode ${item.attributes.chapter}`;
```
ronny1982, line 97: Based on the type definition the volume is always present, so may always be added to the title.

ronny1982 (approved):

web/src/engine/websites/NamiComi.ts
MikeZeDev: Yes. Lang is defined in chapters. You can get all languages from the same manga page by clicking on a button (the api returns all chapters tho). I think the en is for the website interface language.

web/src/engine/websites/NamiComi.ts
MikeZeDev: ok

web/src/engine/websites/NamiComi.ts
MikeZeDev: iirc i had errors when trying to fetch all chapters at once but ill try again

web/src/engine/websites/NamiComi.ts
MikeZeDev: Indeed, volume is nullable.

web/src/engine/websites/NamiComi.ts
MikeZeDev: This has nothing to do with manga language, its the interface. But yeah, just saw website settings, there are others.

PR #612, opened 2024-05-16
Anchira : fix getting pages

MikeZeDev:


web/src/engine/websites/Anchira.ts
```diff
@@ -15,7 +15,9 @@ type APIManga = {
     key: string,
     title: string
     hash?: string,
-    names?: string[]
+    data?: {
```
ronny1982, line 18: Avoid nullable properties and make separate types instead.
Otherwise it will lead to good-like types which are hard to read, understand and error prone with dozens of `if(data.field !== null)` checks
```
type MyAllInOneType = {
    id: ...,
    a?: ...,
    b?: ...,
    c?: ...,
    d?: ...,
    e?: ...,
    f?: ...,
    h?: ...,
    i?: ...,
}
```

PR #616, opened 2024-05-18
Add Toomics template & similar websites

MikeZeDev:


erictbar:
Downloaded a bunch of chapters with VIP subscription and it looks good.

Though, their series that are only available in their new store, which i have purchased some, aren't showing. the store is for series they sublicensed from other providers and aren't in the subscription.

URL of an owned chapter that is not included with subscription: https://toomics.com/en/webtoon/detail/code/221441/ep/4/toon/8031

erictbar:
A specific workaround must be used in order to download purchased chapters from [store (NSFW Link)](https://toomics.com/en/store) exclusive series.

1. While logged in, Paste URL (NSFW link): https://global.toomics.com/en/webtoon/episode/toon/8030
2. Referesh media items. Note that none appear 0/0
3. Logout of Toomics inside the Haruneko browser and disable Family Safe mode
4. Refresh media items, all chapters appear but only free chapters may be downloaded
5. Login with Toomics inside the Haruneko browser
6. Purchased chapters can now be downloaded.

MikeZeDev:
Its possible that the +18 flag for logged users is set differently than the +18 flag for non logged users. 
Its possible that more coding is required. You know what to do :)

PR #619, opened 2024-05-18
Add Dokuha

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/6810


web/src/engine/websites/Dokuha.ts
```diff
@@ -0,0 +1,70 @@
+﻿import { Tags } from '../Tags';
+import icon from './Dokuha.webp';
+import { type Chapter, DecoratableMangaScraper, Page } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import { FetchJSON, FetchWindowScript } from '../platform/FetchProvider';
+import type { Priority } from '../taskpool/DeferredTask';
+
+type JSONImage = {
```
ronny1982, line 8: From the implementation it seems image can be falsy, therefore it should be `undefined | { ... }`

web/src/engine/websites/Dokuha.ts
```diff
@@ -0,0 +1,70 @@
...
+            chunks.push(data.slice(i, i + BYTE_BLOCK));
+        }
+
+        const { decodeRecipes, size } = await FetchJSON<DecryptingData>(new Request(jsonUrl));
+        const result = [];
+        for (const recipe of decodeRecipes) {
+            result[recipe.before] = chunks[recipe.after];
+        }
+
+        return await new Blob(result).slice(0, size).arrayBuffer();
```
ronny1982, line 67: Is this `slice` necessary?

web/src/engine/websites/Dokuha.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const images: JSONImage[] = await FetchWindowScript(new Request(new URL(chapter.Identifier, this.URI)), 'images', 2500);
+        return images.filter(image => image).map(image => new Page(this, chapter, new URL(image.img), { scrambleJSONUrl: image.rectSrc }));
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        const blob = await Common.FetchImageAjax.call(this, page, priority, signal);
+        const data = new Uint8Array(await blob.arrayBuffer());
+        const decrypted = await this.DecryptPicture(data, page.Parameters['scrambleJSONUrl'] as string);
```
ronny1982, line 49: May use `blob.type` as parameter

web/src/engine/websites/Dokuha.ts
ronny1982, line 67: May create a blob of the corresponding type to what `data` was fetched

web/src/engine/websites/Dokuha.ts
```diff
@@ -0,0 +1,70 @@
...
+        return Common.GetTypedData(decrypted);
```
ronny1982, line 50: Maybe redundant if the decryption method would return a blob of the correct type

ronny1982 (approved):

web/src/engine/websites/Dokuha.ts
MikeZeDev: Yes. Its in the original code. Its because they send files with size rounded to multiple of 2k (their chunk making code is not very efficient). Its the reason they have `size` in their json.

web/src/engine/websites/Dokuha.ts
MikeZeDev: those are .bin files. Not sure the server is using proper jpee type.

web/src/engine/websites/Dokuha.ts
MikeZeDev: See previous answer. Pretty sure website send 'applicatoon/octect-stream' as we gather bin files.

web/src/engine/websites/Dokuha.ts
MikeZeDev: See previous answers :D

PR #623, opened 2024-05-23
Port Baozimh

MikeZeDev:


web/src/engine/websites/Baozimh.ts
```diff
@@ -0,0 +1,45 @@
...
+
+function ChapterInfoExtractor(anchor: HTMLAnchorElement) {
+    return {
+        id: anchor.pathname + anchor.search,
+        title: anchor.text.trim()
+    };
+}
+
+@Common.MangaCSS(/^{origin}\/comic\/[^/]+/, '.comics-detail__title')
+@Common.MangasNotSupported()
```
ronny1982, line 15: I think there is a useable list
https://www.baozimh.com/api/bzmhq/amp_comic_list?type=all&region=all&filter=*&page=1&limit=50

web/src/engine/websites/Baozimh.ts
```diff
@@ -0,0 +1,45 @@
...
+        let uri = new URL(chapter.Identifier, this.URI);
+        const sectionSlot = uri.searchParams.get('section_slot');
+        const chapterSlot = uri.searchParams.get('chapter_slot');
+        const linkRegex = new RegExp(`/${sectionSlot}_${chapterSlot}_\\d+\\.html?$`, 'i');
+        for (let run = true; run;) {
+            const data = await FetchHTML(new Request(uri));
+            const pages = [...data.querySelectorAll('.comic-contain amp-img.comic-contain__item')];
+            pagesList.push(...pages.map(element => new Page(this, chapter, new URL(element.getAttribute('src'), this.URI))));
+            uri = new URL([...data.querySelectorAll<HTMLAnchorElement>('div.comic-chapter div.next_chapter a')].pop()?.pathname, this.URI);
+            run = uri && linkRegex.test(uri.href);
```
ronny1982, line 39: The variable `uri` can never be falsy, therefore the check is redundant

web/src/engine/websites/Baozimh.ts
```diff
@@ -0,0 +1,45 @@
...
+        }
+        return pagesList.filter((page, index) => { //not sure if needed anymore
+            return index === pagesList.findIndex(item => item === page);
```
ronny1982, line 42: This will not work as each page is a new instance and therefore reference comparison will never match.
Either compare by value (e.g. `page.Link.href`) or remove this filter.

ronny1982 (approved):

web/src/engine/websites/Baozimh.ts
MikeZeDev: Nice catch !

PR #628, opened 2024-05-30
Add ScanVFOrg

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7119

Thats a bit ugly. 

1) There is one html page per chapter page. Nothing unusual but
2) the "chapters" are Volumes. Meaning, to gather a list of pages for a "chapter", we have to perform more than 100 requests.
3) And then, the same number or request to fetch the images ofc.

Blame the website awful coding i guess.  I wonder if using a pool would be adapted.


web/src/engine/websites/ScanVFOrg.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    public constructor() {
+        super('scanvforg', 'ScanVF.org', 'https://scanvf.org', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Language.French, Tags.Source.Scanlator);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
```
ronny1982, line 25: May consider using partitioning algorithm such as divide and conquer to perform head requests and find the response that was redirected to url with parameter `bypass=1`, which would be the last page number.
`fetch('https://scanvf.org/scan/705/170', { method: 'HEAD' });`

Than generate all page links and implement `FetchImage` accordingly to extract the image from the HTML page.

ronny1982 (changes requested):
Should optimize `FetchPages` as it is supposed to be a fast operation.

PR #637, opened 2024-06-05
Block Bloatware URL Patters in Electron Client

ronny1982:


app/electron/src/ipc/BloatGuard_test.ts
```diff
@@ -0,0 +1,34 @@
...
+import { BloatGuard } from './BloatGuard';
+
+class TestFixture {
+
+    public readonly mockIPC = {
+        Listen: vi.fn(),
+    } as unknown as IPC<never, never>;
+
+    private readonly mockWebContents = {
+        getURL: vi.fn(),
```
ronny1982, line 14: Unused declaration

ronny1982: Addressed in 43c423b9

PR #640, opened 2024-06-08
Add MangaRaw1001

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7148


web/src/engine/websites/MangaRaw1001.ts
```diff
@@ -0,0 +1,40 @@
+import { Tags } from '../Tags';
+import icon from './MangaRaw1001.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import * as MangaReader from './decorators/MangaReaderCMS';
+
+const pageScript = `
```
ronny1982, line 7: I can't see anything special in this script. Any reason why using `FetchBrowser` instead of implementing `FetchPages` based on the normal `Fetch`?

ronny1982 (approved):

web/src/engine/websites/MangaRaw1001.ts
MikeZeDev: FetchBrowser? You mean because i use FetchWindowScript? 

Its just because we need the CHAPTER_ID variable for the ajax request. Which is only available inside chapter page.
It cant be inferred from resource uri.

I find it cleaner to do everything in one go than using FetchWindowScript or Regexp to get CHAPTER_ID  then perform the ajax request ourselves.

ronny1982: It seemed to me like the chapter ID was in the URL, but then this assumption was probably wrong
image

ronny1982: Note: It seems the real chapter ID is available in the JSON chapter list on the manga page

image

MikeZeDev: Nice catch, now i wonder if this ld-json thing is reliable.

PR #647, opened 2024-06-15
Add HentaiForce, 3Hentai

MikeZeDev:
Similar to NHentai


web/src/engine/websites/HentaiForce.ts
```diff
@@ -0,0 +1,24 @@
+import { Tags } from '../Tags';
+import icon from './HentaiForce.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+
+export function ExtractImageLink(element: HTMLImageElement) {
```
ronny1982, line 6: Avoid exporting internal functions.

web/src/engine/websites/ThreeHentai.ts
```diff
@@ -0,0 +1,21 @@
+import { Tags } from '../Tags';
+import icon from './ThreeHentai.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import { ExtractImageLink } from './NHentai';
```
ronny1982, line 5: Are these sites related, so it i safe to assume they share the same business logic?
Otherwise changing this method may unintentionally break this website in the future.

web/src/engine/websites/ThreeHentai.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@Common.MangaCSS(/^https:\/\/(es\.|fra\.|it\.|pt\.|ru\.)?3hentai\.net\/d\/\d+/, 'div#main-info span.middle-title')
```
ronny1982, line 7: Maybe use a more generic regex such as `[a-z]{2,3}\.` instead of listing all country codes

ronny1982 (approved):

PR #648, opened 2024-06-16
Port ScanManga

MikeZeDev:
nasty website, with a weird way to fetch images.


web/src/engine/websites/ScanManga.ts
```diff
@@ -0,0 +1,78 @@
+import { Tags } from '../Tags';
+import icon from './ScanManga.webp';
+import { DecoratableMangaScraper, type Page } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import type { Priority } from '../taskpool/DeferredTask';
+import { Fetch } from '../platform/FetchProvider';
+
+const pagescript = `
+    new Promise( resolve => {
+        fetch("/api/lel/" + idc + '.json', {
```
ronny1982, line 10: Consider using async/await instead of promise chaining

web/src/engine/websites/ScanManga.ts
```diff
@@ -0,0 +1,78 @@
...
+                const _page = _CHAPTERDATA.p[_ITEM];
+                let _path_ = _BASEURL;
+                const _extension = _page.e;
+                const finalpath = "/" + encodeURIComponent(_page.f) + '.' + _extension + '';
+                _path_ += finalpath;
+                return _path_;
+            });
+
+            resolve(_IMGLIST);
+        });
```
ronny1982, line 40: Catch is missing, in case of an error the promise will never resolve or reject

web/src/engine/websites/ScanManga.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        return this.imageTaskPool.Add(async () => {
+            const request = new Request(page.Link, {
+                credentials: 'omit',
```
ronny1982, line 63: Are these specific `credentials`, `referrerPolicy` and `mode` really necessary?

ronny1982 (changes requested):

web/src/engine/websites/ScanManga.ts
MikeZeDev: Yes. website is very picky about fetching pages.

PR #649, opened 2024-06-16
Add MangaRawAC

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7190

(expand SpoilerPlus)


web/src/engine/websites/MangaRawAC_e2e.ts
```diff
@@ -0,0 +1,26 @@
+﻿import { describe } from 'vitest';
+import { TestFixture } from '../../../test/WebsitesFixture';
+
+const config = {
+    plugin: {
+        id: 'mangarawac',
+        title: 'MangaRawAC'
+    },
+    container: {
+        url: 'https://mangaraw.ac/manga/%e5%9c%b0%e7%90%83%e9%98%b2%e8%a1%9b%e9%9a%8aX-raw-free/',
```
ronny1982, line 10: I think encoding the URLs would be more readable `encodeURI('https://mangaraw.ac/manga/地球防衛隊X-raw-free/')`

ronny1982 (approved):

web/src/engine/websites/MangaRawAC_e2e.ts
ronny1982: There are similar tests such as Doujin69 which seems to work fine.
Pasting in VS Code should be no problem, it is likely the Browser that performs the encoding when copying the URL from the address bar.

MikeZeDev: I copied it for the browser console output when doing decodeURI(location.href). It works flawlessly for the manga URL/ID but fails unsuspectingly for the chapter id. I dont really get it. 

If i encodeuri the pathname in chapterExtractor and put a decodeURI in test , the test works but its ugly AF.

ronny1982: The `encodeURI` method uses uppercase encoding, while the website uses lowercase encoding. Checking the chapter links on the website it can be seen that the manga identifier is lowercase encoded while the chapter part is not encoded at all. HakuNeko will encode the chapter part uppercase. Now the chapter identifier is inconsistently mixed with lowercase and uppercase encoding.

image

Possible solutions would be to keep the encoded URLs like you have done initially, or explicitly use lowercase/uppercase encodings.

```typescript
const mangaID = encodeURI('/manga/呪術廻戦-raw-free/').toLowerCase();
const config = {
    plugin: {
        id: 'mangarawac',
        title: 'MangaRawAC'
    },
    container: {
        url: 'https://mangaraw.ac' + mangaID,
        id: mangaID,
        title: '呪術廻戦'
    },
    child: {
        id: mangaID + encodeURI('第1話/'),
        title: '第1話'
    },
    entry: {
        index: 0,
        size: 156_354,
        type: 'image/jpeg'
    }
};
```

MikeZeDev: I did see the mixed case problem yup, i guess i was too dumb to find a fix :D

PR #652, opened 2024-06-17
MangaDistrict: fix image attr

AwkwardPeak7:


web/src/engine/websites/MangaDistrict.ts
```diff
@@ -4,10 +4,14 @@ import { DecoratableMangaScraper } from '../providers/MangaPlugin';
...
 
+const script = `
+    const images = [...document.querySelectorAll('div.page-break img')];
+    images.map(image => image.dataset['wpfcOriginalSrc'] || image.src);
+`
 @Madara.MangaCSS(/^{origin}\/read-scan\/[^/]+\/$/, 'meta[property="og:title"]:not([content*="MANGA DISTRICT"])')
 @Madara.MangasMultiPageAJAX()
 @Madara.ChaptersSinglePageAJAXv2()
-@Madara.PagesSinglePageCSS()
+@Common.PagesSinglePageJS(script, 1000)
```
MikeZeDev, line 14: Suggestion : While it works, we should avoid injecting javascript when we can.

Please consider using 
`@Common.PagesSinglePageCSS(query, PageExtractor)` and define a `PageExtractor `that return either image.dataset.wpfcOriginalSrc either image.src :)

Like in : https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/ComicOnlineFree.ts#L11

MikeZeDev (changes requested):

MikeZeDev (approved):
Thanks for your contribution.

PR #656, opened 2024-06-22
Add Mangareaderto

MikeZeDev:
Tests are covering Volume, Chapters, vanilla and scrambled pictures.


web/src/engine/websites/MangaReaderTo.ts
```diff
@@ -0,0 +1,282 @@
...
+        const uri = new URL(`ajax/image/list/${chapterurl.href.includes('chapter') ? 'chap' : 'vol'}/${readingId}?quality=high`, this.URI);
+        const { status, html } = await FetchJSON<AJAXResponse>(new Request(uri));
+        if (!status) return [];
+
+        const dom = new DOMParser().parseFromString(html, 'text/html');
+        const imagesArr = Array.from(dom.querySelectorAll<HTMLDivElement>('div.iv-card'));
+        return imagesArr.map(image => new Page(this, chapter, new URL(image.dataset.url, this.URI), { shuffled: [...image.classList].includes('shuffled') }));
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
```
ronny1982, line 68: The complexity of this method is t high, consider extracting parts of it into separate functions.
It is also recommend to place functions which are copied directly from the website script outside of the scraper class.

web/src/engine/websites/MangaReaderTo.ts
```diff
@@ -0,0 +1,282 @@
...
+        let blob = await Common.FetchImageAjax.call(this, page, priority, signal);
```
ronny1982, line 69: Prefer `const` over `let`

web/src/engine/websites/MangaReaderTo.ts
```diff
@@ -0,0 +1,282 @@
...
+        return !page.Parameters['shuffled'] ? blob : DeScramble(blob, async (image, ctx) => {
+
+            const pieceDimension = 200;
+            const defaultSeed = 'stay';
+            const numpieces = Math.ceil(image.width / pieceDimension) * Math.ceil(image.height / pieceDimension);
```
ronny1982, line 74: May use more expressive names such as tiles or blocks instead of pieces

web/src/engine/websites/MangaReaderTo.ts
```diff
@@ -0,0 +1,282 @@
...
+            const numcol = Math.ceil(image.width / pieceDimension);
```
ronny1982, line 75: Avoid shortening variable names `numberColumns` is more readable than `numcol`

web/src/engine/websites/MangaReaderTo.ts
```diff
@@ -0,0 +1,282 @@
...
+            const slicesMap = new Map<string, TSlice[]>();
+            for (let index = 0; index < numpieces; index++) {
+                const g = Math.trunc(index / numcol);
```
ronny1982, line 78: Use more expressive variable name instead of `g` (i assume it is the row index/number?)

web/src/engine/websites/MangaReaderTo.ts
```diff
@@ -0,0 +1,282 @@
...
+                const slice: TSlice = { x: (index - g * numcol) * pieceDimension, y: g * pieceDimension, width: undefined, height: undefined };
+                slice.width = pieceDimension - (slice.x + pieceDimension <= image.width ? 0 : slice.x + pieceDimension - image.width);
```
ronny1982, line 80: Simplify the whole expression with`Math.min` or `Math.max` for better readability.
I think it is same as `Math.min(image.width - slice.x, pieceDimension)`.
Same for height calculation.

ronny1982 (approved):
I didn't review all the declared function inside of `FetchImage` in detail as i believe these are copy/pasted from the website script.

web/src/engine/websites/MangaReaderTo.ts
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const chapterurl = new URL(chapter.Identifier, this.URI);
+        const [data] = await FetchCSS(new Request(chapterurl), 'div#wrapper');
+        const readingId = data.dataset.readingId;
+
+        const uri = new URL(`ajax/image/list/${chapterurl.href.includes('chapter') ? 'chap' : 'vol'}/${readingId}?quality=high`, this.URI);
+        const { status, html } = await FetchJSON<AJAXResponse>(new Request(uri));
+        if (!status) return [];
```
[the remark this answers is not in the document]
MikeZeDev: status is a boolean value, is my check incorrect? 
In the unexpected case it would be undefined or null, wouldnt my check be fine?

web/src/engine/websites/MangaReaderTo.ts
MikeZeDev: oops

web/src/engine/websites/MangaReaderTo.ts
MikeZeDev: I didnt dig much into the code. It was heavily obfuscated :D

web/src/engine/websites/MangaReaderTo.ts
MikeZeDev: Nice catch!

PR #657, opened 2024-06-24
Add ColoredManga

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7211


web/src/engine/websites/ColoredManga.ts
```diff
@@ -0,0 +1,96 @@
...
+        return chapters.map(chapter => new Chapter(this, manga, chapter.id, [chapter.number, chapter.title].join(' - ').trim().replace(/-$/, '')));
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const pageList: Page[] = [];
+        const { name, chapters } = await this.GetMangaData(chapter.Parent.Identifier);
+        const url = new URL(`/images/content/${name}/${chapter.Title}`, this.URI);
+        const chapterItem = chapters.find(item => item.id === chapter.Identifier);
+        for (let i = 1; i < chapterItem.totalImage; i++) {
+            const page = new Page(this, chapter, url, { number: i.toString().padStart(4, '0') });
```
ronny1982, line 61: Consider to provide the `path` as page parameter instead as part of `url`

ronny1982 (commented):
Review postponed due to website currently unreachable

web/src/engine/websites/ColoredManga.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const mangas = await Common.FetchMangasSinglePageCSS.call(this, provider, '/manga', 'div#themes_outside__WCut6 a:not([id])');
+        return mangas.map(manga => new Manga(this, provider, manga.Identifier.match(/\/manga\/([^/]+)/)[1], manga.Title));
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const { chapters } = await this.GetMangaData(manga.Identifier);
+        return chapters.map(chapter => new Chapter(this, manga, chapter.id, [chapter.number, chapter.title].join(' - ').trim().replace(/-$/, '')));
```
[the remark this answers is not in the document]
MikeZeDev: Got to be extra careful, as this "title" is used for image path tho.

web/src/engine/websites/ColoredManga.ts
```diff
@@ -0,0 +1,98 @@
...
+                chapter.title ? '-' : '',
+                chapter.title,
+            ].join(' ').trim();
+            return new Chapter(this, manga, chapter.id, title);
+        });
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const { name, chapters } = await this.GetMangaData(chapter.Parent.Identifier);
+        const url = new URL(`/images/content/${name}/${chapter.Title}`, this.URI);
```
ronny1982, line 64: Composing an URL to get the encoded pathname just to later decode the pathname again seems redundant.
Consider using the pathname directly:
```javascript
const path = `/images/content/${name}/${chapter.Title}`;
```

web/src/engine/websites/ColoredManga.ts
```diff
@@ -0,0 +1,98 @@
...
+        const { totalImage } = chapters.find(item => item.id === chapter.Identifier);
+        return new Array(totalImage).fill(0).map((_, index) => new Page(this, chapter, this.URI, { path: url.pathname, number: (index+1).toString().padStart(4, '0') }));
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        return await this.imageTaskPool.Add(async () => {
+
+            const formdata = new FormData();
+            formdata.append('path', decodeURI(page.Parameters['path'] as string));
```
ronny1982, line 73: No need to decode in case of applying the previous suggestion

web/src/engine/websites/ColoredManga.ts
```diff
@@ -0,0 +1,98 @@
...
+
+type APIChapter = {
+    id: string,
+    title: string,
+    number: string,
+    totalImage: number
+}
+
+type APIImage = {
+    image : string
```
ronny1982, line 21: Document that this string is actually a data uri (image data uri)

web/src/engine/websites/ColoredManga.ts
```diff
@@ -0,0 +1,98 @@
...
+        return await this.imageTaskPool.Add(async () => {
+
+            const formdata = new FormData();
+            formdata.append('path', decodeURI(page.Parameters['path'] as string));
+            formdata.append('number', page.Parameters['number'] as string);
+
+            const request = this.CreateRequest('dynamicImages', formdata);
+            const { image } = await FetchJSON<APIImage>(request);
+
+            const response = await fetch(image);//turn base64image into response
```
ronny1982, line 79: Remove comment and document it on the type declaration

web/src/engine/websites/ColoredManga.ts
```diff
@@ -0,0 +1,98 @@
...
+        }, priority, signal);
+    }
+
+    private async GetMangaData(mangaSlug: string): Promise<APIManga> {
+        const formdata = new FormData();
+        formdata.append('id', mangaSlug);
+        return await FetchJSON<APIManga>(this.CreateRequest('selectedManga', formdata));
+    }
+
+    private CreateRequest(endpoint: string, formData: FormData): Request {
```
ronny1982, line 91: Instead of accepting `FormData`, consider using `Record<string, string>` and then compose the form inside of this method (by looping through all keys of the record)

ronny1982 (approved):

MikeZeDev:
Turn out there are manga where chapters are not fetched because they are scattered inside an array of volumes in JSON.
Gotta make a big fix for that.

web/src/engine/websites/ColoredManga.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        return await this.imageTaskPool.Add(async () => {
+
+            const request = this.CreateRequest('dynamicImages', {
+                path: page.Parameters['path'] as string,
+                number: page.Parameters['number'] as string
+            });
+
+            const { image } = await FetchJSON<APIImage>(request);
+            const response = await fetch(image);
```
ronny1982, line 100: Will this work for `image` > 2MB?
There were some issues in HakuNeko about this:
- https://github.com/manga-download/hakuneko/pull/4336#discussion_r791682275
- https://github.com/manga-download/hakuneko/pull/4783#discussion_r912200805

MikeZeDev: Is the limitation inherent to the fetch api or for data uri? 
In the first case we can convert it by ourselves , in the second case its the website problem 🤷

MikeZeDev: [Length limitations](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs#length_limitations)

> Browsers are not required to support any particular maximum length of data. For example, the Opera 11 browser limited URLs to 65535 characters long which limits data URLs to 65529 characters (65529 characters being the length of the encoded data, not the source, if you use the plain data:, without specifying a MIME type). Firefox version 97 and newer supports data URLs of up to **32MB** (before 97 the limit was close to 256MB). Chromium objects to URLs over **512MB**, and Webkit (Safari) to URLs over **2048MB**.

Shouldnt it be fine?

MikeZeDev: Tested with a 5mb dataURI 👍 

image

ronny1982: Was this tested with both, the `NW.js` client and the `Electron` client?

MikeZeDev: > Was this tested with both, the `NW.js` client and the `Electron` client?

Affirmative. Event loaded a bunch of them . All 5.5 mb pictures (so even more larger data URI)

image

PR #663, opened 2024-07-04
Add ProjectSuki

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/3294


web/src/engine/websites/ProjectSuki.ts
```diff
@@ -0,0 +1,72 @@
...
+
+                    const images = [];
+                    images.push(...[...document.querySelectorAll('img.img-fluid')].map(image => image.src));
+
+                    const dom = new DOMParser().parseFromString(jsondata.src, 'text/html');
+                    images.push(...[...dom.querySelectorAll('img.img-fluid')].map(image => image.src));
+                    resolve(images);
+
+                } else {
+                    throw Error('Cant get images :/ !');
```
ronny1982, line 33: Do not throw in asynchronous function, this will not be handled by promise. use `reject` instead.

web/src/engine/websites/ProjectSuki.ts
```diff
@@ -0,0 +1,72 @@
...
+            const chapterid = window.location.href.split('/') [5];
+            const body = JSON.stringify({
+                bookid: bookid,
+                chapterid: chapterid,
+                first: true
+            });
+
+            const xhr = new XMLHttpRequest();
+            xhr.onload = function () {
+                if (xhr.status >= 200 && xhr.status < 300) {
```
ronny1982, line 22: Use `try ... catch` in asynchronous function and properly reject the promise to avoid uncaught exceptions which will lead to the promise never being fulfilled.

web/src/engine/websites/ProjectSuki.ts
```diff
@@ -0,0 +1,72 @@
...
+import { DecoratableMangaScraper, Manga, type MangaPlugin } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import { FetchWindowScript } from '../platform/FetchProvider';
+
+const pageScript = `
+    new Promise ((resolve, reject) => {
+        try {
+
+            const element = document.querySelector('.strip-reader');
+            const bookid = window.location.href.split('/') [4];
```
ronny1982, line 12: Consider use of spreading `const [, , , , bookid, chapterid ] = window.location.href.split('/');`

web/src/engine/websites/ProjectSuki.ts
```diff
@@ -0,0 +1,72 @@
...
+            const chapterid = window.location.href.split('/') [5];
+            const body = JSON.stringify({
+                bookid: bookid,
```
ronny1982, line 15: Can be simplified since JSON field and variable name are identical `{ bookid, chapterid, first: true }`

web/src/engine/websites/ProjectSuki.ts
```diff
@@ -0,0 +1,72 @@
...
+                first: true
+            });
+
+            const xhr = new XMLHttpRequest();
+            xhr.onload = function () {
+                if (xhr.status >= 200 && xhr.status < 300) {
+                    const jsondata =  JSON.parse(xhr.response);
+
+                    const images = [];
+                    images.push(...[...document.querySelectorAll('img.img-fluid')].map(image => image.src));
```
ronny1982, line 26: Can be simplified by assigning the entries directly instead of pushing into an empty array `const images = [...document/*...*/].map(/*...*/);`

ronny1982 (changes requested):

web/src/engine/websites/ProjectSuki_e2e.ts
```diff
@@ -0,0 +1,26 @@
...
+        id: '/book/207989',
+        title: 'Absolute Sword Sense'
+    },
+    child: {
+        id: '/read/207989/34655/1',
+        title: 'Chapter 89'
+    },
+    entry: {
+        index: 1,
+        size: 2_493_672,
```
ronny1982, line 20: Any image in this chapter with smaller size?

ronny1982 (approved):

PR #673, opened 2024-07-11
Add Akuma.moe

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/3870


web/src/engine/websites/Akuma.ts
```diff
@@ -0,0 +1,47 @@
+import { Tags } from '../Tags';
+import icon from './Akuma.webp';
+import { Chapter, DecoratableMangaScraper, type Manga } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+
+const pageScript = `
+    new Promise( (resolve, reject) => {
+        try {
```
ronny1982, line 8: The `try ... catch` is redundant in synchronous context scope.

web/src/engine/websites/Akuma.ts
```diff
@@ -0,0 +1,47 @@
...
+                headers: {
+                    'X-CSRF-TOKEN': token,
+                    'X-Requested-With': 'XMLHttpRequest',
+
+                },
+                 origin: window.location.origin,
+                 referrer: window.location.href,
+                 method: 'POST',
+            }).then(response => response.json()
+                .then(data => resolve(data.map(image => img_prt + '/' + image))));
```
ronny1982, line 22: Either use the `.catch()` fluent method to reject the promise, or better use `async ... await` for `fetch`.

web/src/engine/websites/Akuma.ts
```diff
@@ -0,0 +1,47 @@
...
+            const url = window.location.href.replace(/\\/1$/, '');
+            const token = document.querySelector('meta[name="csrf-token"]').content.trim();
+            fetch(url, {
+                credentials: 'include',
+                headers: {
+                    'X-CSRF-TOKEN': token,
+                    'X-Requested-With': 'XMLHttpRequest',
+
+                },
+                 origin: window.location.origin,
```
ronny1982, line 18: Redundant, this is default behavior of `fetch`

web/src/engine/websites/Akuma.ts
```diff
@@ -0,0 +1,47 @@
...
+                 referrer: window.location.href,
```
ronny1982, line 19: Redundant, this is default behavior of `fetch`

web/src/engine/websites/Akuma.ts
```diff
@@ -0,0 +1,47 @@
...
+
+        } catch (error) {
+            reject(error);
+        }
+    });
+ `;
+
+@Common.MangaCSS(/^{origin}\/g\/[^/]+$/, 'h1.entry-title')
+@Common.MangasNotSupported()
+@Common.PagesSinglePageJS(pageScript)
```
ronny1982, line 32: Maybe extracting the token in `Initialize()` and store it as field, so the JSON API can be accessed directly when requesting pages.

web/src/engine/websites/Akuma.ts
```diff
@@ -0,0 +1,47 @@
...
+    public constructor() {
+        super('akuma', 'Akuma.moe', 'https://akuma.moe', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Media.Manga, Tags.Language.Multilingual, Tags.Source.Aggregator, Tags.Rating.Pornographic);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        return [new Chapter(this, manga, new URL(`${manga.Identifier}/1`, this.URI).pathname, manga.Title)];
```
ronny1982, line 45: Identifier seems to be semantically wrong, it points to the first page of the 'chapter.
Since there is no real chapter it should be consistent and use the manga identifier instead?

web/src/engine/websites/Akuma_e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+        id: '/g/dey7wcbd/1',
+        title: '[Isikenpi] Lethargic Sister Tolerates her Brothers Pranks [English][DarklordMTLs]'
+    },
+    entry: {
+        index: 5,
+        size: 223_038,
+        type: 'image/jpeg'
+    }
+});
+describe(fixture.Name, async () => (await fixture.Connect()).AssertWebsite());
```
ronny1982, line 24: Website is protected by DDoS Guard for me, is this correctly bypassed in this test?

ronny1982 (approved):

web/src/engine/websites/Akuma.ts
MikeZeDev: i use fetch.then, and its sync? gosh i will never understand

web/src/engine/websites/Akuma.ts
MikeZeDev: I'll still need to get the javascript variable of the CDN in chapter page so i dont think its worth.

web/src/engine/websites/Akuma_e2e.ts
MikeZeDev: I wont let a test that is not passed uncommented. Ofc its passing ;)

web/src/engine/websites/Akuma.ts
MikeZeDev: The script to get pages has to be executed on this url. 

Had i used the common decorator (ChapterSinglePageCSS), the "chapter" link title would have been "Start reading" with no **clean** way to get the title in chapter extractor.

In any case i would need to override either FetchPages (to put the /1), either FetchChapters.

PR #677, opened 2024-07-15
Add Nico nico seiga

MikeZeDev:


web/src/engine/websites/NicoNicoSeiga.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const { meta, data: { result } } = await FetchJSON<APIResult<APIPage[]>>(new Request(new URL(`episodes/${chapter.Identifier}/frames`, this.apiUrl)));
+        return meta.status === 200 ? result.map(page => new Page(this, chapter, new URL(page.meta.source_url), { drm_hash: page.meta.drm_hash })) : [];
+    }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        const data = await Common.FetchImageAjax.call(this, page, priority, signal);
+        const encrypted = await new Response(data).arrayBuffer();
```
ronny1982, line 84: Use `data.arrayBuffer()` instead

web/src/engine/websites/NicoNicoSeiga.ts
```diff
@@ -0,0 +1,96 @@
...
+        const key = page.Parameters['drm_hash'] as string;
+        const decrypted = this.Decrypt(new Uint8Array(encrypted), key);
+        return Common.GetTypedData(Buffer.from(decrypted));
```
ronny1982, line 87: Is it possible to use `return new Blob(decrypted, { type: data.type })` instead?

web/src/engine/websites/NicoNicoSeiga.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private Decrypt(buffer: Uint8Array, key: string): Uint8Array {
```
ronny1982, line 90: This seems to mutate the input parameter `buffer`, which works for this use case but is unexpected behavior.
Either document this behavior, or return a copy with the decrypted data (e.g., `map` the xored buffer).

ronny1982 (approved):

web/src/engine/websites/NicoNicoSeiga.ts
```diff
@@ -0,0 +1,96 @@
...
+    public override get Icon() {
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return this.mangaRegexp.test(url);
+    }
+
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const title = (await FetchCSS<HTMLHeadingElement>(new Request(url), 'section.comic_header h2.title')).pop().textContent.trim();
```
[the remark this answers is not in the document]
MikeZeDev: There is only one anyway

web/src/engine/websites/NicoNicoSeiga.ts
MikeZeDev: Nope. 
image

web/src/engine/websites/NicoNicoSeiga.ts
MikeZeDev: Like its done for Kadocomi? 

https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/KadoComi.ts#L143

By the way, shouldn't GetTypedData take an Uint8Array as parameter instead of an ArrayBuffer? Isn't ArrayBuffer too generic? What if someone use Uint16Array?

PR #678, opened 2024-07-17
Copymanga : fix getting chapters and pages

MikeZeDev:


web/src/engine/websites/CopyManga.ts
```diff
@@ -104,47 +87,50 @@ export default class extends DecoratableMangaScraper {
...
-        return orderedPages.map(page => new Page(this, chapter, new URL(page)));
+
+        const [dataElement] = await FetchCSS<HTMLDivElement>(new Request(new URL(`/comic/${chapter.Parent.Identifier}/chapter/${chapter.Identifier}`, this.URI)), '.imageData');
+        const imageData = dataElement.getAttribute('contentKey');
+        const images = await this.Decrypt<APIPage[]>(imageData);
+        return images.map(image => new Page(this, chapter, new URL(image.url)));
+
+    }
+
+    private async Decrypt<T>(encryptedData: string, keyString: string = this.defaultKey): Promise<T> {
```
ronny1982, line 106: Consider to remove the `keyString` parameter as this is never provided in any call.

web/src/engine/websites/CopyManga.ts
```diff
@@ -104,47 +87,50 @@ export default class extends DecoratableMangaScraper {
...
+
+        const key = Buffer.from(keyString, 'utf-8');
+        const iv = Buffer.from(encryptedData.substring(0x0, 0x10), 'utf-8');
```
ronny1982, line 109: Personally in this case i prefer to use `0` and `16` instead of their hex representation.
Same for `cipher`.

web/src/engine/websites/CopyManga.ts
```diff
@@ -104,47 +87,50 @@ export default class extends DecoratableMangaScraper {
...
+        const cipher = Buffer.from(encryptedData.substring(0x10, encryptedData.length), 'hex');
+
+        const secretKey = await crypto.subtle.importKey(
```
ronny1982, line 112: Inconsistent line breaks for short parameters, either break no short parameters or all
```typescript
await crypto.subtle.importKey('raw', key, {
      name: 'AES-CBC',
      length: 128
  }, true, ['encrypt', 'decrypt']);
```

web/src/engine/websites/CopyManga.ts
```diff
@@ -104,47 +87,50 @@ export default class extends DecoratableMangaScraper {
...
+            'raw',
+            key,
+            {
+                name: 'AES-CBC',
+                length: 128
+            }, true, ['encrypt', 'decrypt']);
```
ronny1982, line 118: Is the `encrypt` capability required?

web/src/engine/websites/CopyManga.ts
```diff
@@ -104,47 +87,50 @@ export default class extends DecoratableMangaScraper {
...
+            name: 'AES-CBC',
+            iv: iv
+        }, secretKey, cipher);
+
+        return JSON.parse(Buffer.from(data).toString('utf-8')) as T;
     }
 
     private CreateApiRequest(pathname: string): Request {
-        const request = new Request(new URL(`/api/v3/${pathname}`, this.apiurl).href, {
+        const request = new Request(new URL(pathname, this.URI), {
```
ronny1982, line 129: No need for intermediate variable `request` can be return directly.

ronny1982 (approved):

web/src/engine/websites/CopyManga.ts
MikeZeDev: Ok, I dont have any preference, its just copypasted anyway .

web/src/engine/websites/CopyManga.ts
MikeZeDev: not really

web/src/engine/websites/CopyManga.ts
ronny1982: Avoid placing 1:1 copy/pasted functions from the website into the class as it may interfere with the `this` scope.
It would be preferred to place them outside of the class.

PR #679, opened 2024-07-18
Add Kumanga

MikeZeDev:
Once you Bypass Cloudflare, it works without problems.


web/src/engine/websites/KuManga.ts
```diff
@@ -0,0 +1,101 @@
...
+        resolve(pUrl.map(page => new URL(page.imgURL, window.location.origin).href))
+    });
+`;
+
+const getTokenScript = `
+    new Promise (resolve => {
+        function btoaReverse(content) {
+            return btoa(content).split('').reverse().join('');
+        }
+        const data = document.querySelector('#searchinput').getAttribute('dt');
```
ronny1982, line 27: Can be wrapped in `const data = btoaReverse(...);`

ronny1982 (approved):

PR #685, opened 2024-07-19
ReadComicOnline: fix getting pages again

MikeZeDev:
fix mangalist too.

Relevant commit : https://github.com/manga-download/haruneko-proprietary/commit/4b5cdcbe31f5751cf4c2c5d1e7eaa2fc36059fa6


ronny1982:
I don't think obfuscation is necessary. RCO (ReadComicOnline) is probably simply reacting to [Tachiyomi's latest fixes](https://github.com/keiyoushi/extensions-source/commits/main/src/en/readcomiconline).
They are probably not watching HaruNeko yet.

MikeZeDev:
> I don't think obfuscation is necessary. RCO (ReadComicOnline) is probably just reacting to [Tachiyomi's recent RCO related fixes](https://github.com/keiyoushi/extensions-source/commits/main/src/en/readcomiconline). They are probably not watching HaruNeko yet.

Yeah, but they seems determinated. Hum, ill let it like that for now.

PR #688, opened 2024-07-22
Add Hitomi

MikeZeDev:


ronny1982:
I cannot access the website, it always times out.

ronny1982:
Seems to be region locked, is accessible from Spain

MikeZeDev:
I have problem accessing it 🤷

ronny1982 (approved):

PR #691, opened 2024-07-22
Detect Flatmanga AntiDdoSS

MikeZeDev:
relevant commit : https://github.com/manga-download/haruneko-proprietary/commit/2f544fd4cd9b27e5b0426af34ae1c561668c1fb6

Dont want to put more code on Flatmanga PR and since its touching obfuscated code for the other repo i think its better done in a separate PR.


MikeZeDev:
> Only generic detections are supposed to be added to `AntiScrapingDetection.ts` Add it to the related file as done in e.g., [MangaLink](https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/MangaLink.ts#L8-L11)

As is it, it would be related to 6 websites, i think its enough. We shouldnt add it to all websites affected (as it will be redundant), and which one should we choose ?

ronny1982:
It must only be added once not six times.
Technically it doesn't matter where it is added, but semantically it should be added in a file that is related to the bypass e.g., The FlatManga decorator file.

PR #695, opened 2024-07-25
ReadComicOnline : fix getting pages

MikeZeDev:
To get pages, the script from Hakuneko cant be used directly. If used directly, we get a random amount of blank gif.
Adding a delay works but i dont feel like adding an arbitrary delay this website is touchy, often has server slowness, etc..

I've had other ideas but they are quite dirty.
I ended up with this method 🤷  If you have better ideas i'll gladly take them

### DIFF
--- a/web/src/engine/websites/ReadComicOnline.ts
+++ b/web/src/engine/websites/ReadComicOnline.ts
@@ -31,7 +31,7 @@ const pageScript = `
                 tries++;
                 if (tries > 10) {
                     clearInterval(interval);
-                    reject('Unable to get pictures after more than 10 tries !');
+                    reject(new Error('Unable to get pictures after more than 10 tries !'));
                 }
             }
         }, 1000);
--- a/web/src/engine/websites/ReadComicOnline.ts
+++ b/web/src/engine/websites/ReadComicOnline.ts
@@ -16,19 +16,18 @@ function ChapterExtractor(anchor: HTMLAnchorElement) {
 
 const pageScript = `
     new Promise( (resolve, reject) => {
-        const arrayName = document.documentElement.innerHTML.match(/(\\w+)\\.push\\(pth/)[1];
         let tries = 0;
         const interval = setInterval(function () {
             try {
-                const imagearray = window[arrayName];
-                if (!Array.isArray(imagearray)) return;
-                const lastElement = imagearray[imagearray.length - 1];
-                const url = !/blank.gif/i.test(lastElement) && new URL(lastElement);
-                if (url) {
+                const links = [ ...document.querySelectorAll('#divImage img') ].map(img => img.src);
+                if(links.length > 0 && !links.some(link => /blank.gif/i.test(link))) {
                     clearInterval(interval);
-                    resolve (imagearray);
+                    resolve(links);
                 }
             } catch (error) {
+                clearInterval(interval);
+                reject(error);
+            } finally {
                 tries++;
                 if (tries > 10) {
                     clearInterval(interval);
--- a/web/src/engine/websites/ReadComicOnline.ts
+++ b/web/src/engine/websites/ReadComicOnline.ts
@@ -2,6 +2,7 @@ import { Tags } from '../Tags';
 import icon from './ReadComicOnline.webp';
 import { DecoratableMangaScraper } from '../providers/MangaPlugin';
 import * as Common from './decorators/Common';
+import { FetchWindowScript } from '../platform/FetchProvider';
 
 function ChapterExtractor(anchor: HTMLAnchorElement) {
     const link = new URL(anchor.href);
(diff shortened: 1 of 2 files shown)


web/src/engine/websites/ReadComicOnline.ts
```diff
@@ -13,19 +14,36 @@ function ChapterExtractor(anchor: HTMLAnchorElement) {
...
+                const imagearray = window[arrayName];
+                if (!Array.isArray(imagearray)) return;
+                const lastElement = imagearray[imagearray.length - 1];
+                const url = !/blank.gif/i.test(lastElement) && new URL(lastElement);
+                if (url) {
+                    clearInterval(interval);
+                    resolve (imagearray);
+                }
+            } catch (error) {
+                tries++;
```
ronny1982, line 32: This catch block will never be reached in case the last image is always a valid URL containing `blank.gif`

web/src/engine/websites/ReadComicOnline.ts
```diff
@@ -13,19 +14,36 @@ function ChapterExtractor(anchor: HTMLAnchorElement) {
...
-}
+const pageScript = `
+    new Promise( (resolve, reject) => {
+        const arrayName = document.documentElement.innerHTML.match(/(\\w+)\\.push\\(pth/)[1];
+        let tries = 0;
+        const interval = setInterval(function () {
+            try {
+                const imagearray = window[arrayName];
+                if (!Array.isArray(imagearray)) return;
+                const lastElement = imagearray[imagearray.length - 1];
```
ronny1982, line 25: Maybe perform check on all elements not just the last one, something like
`imagearray.filter(link => !/blank.gif/i.test(link)).map(link => new URL(link))`

ronny1982 (changes requested):

web/src/engine/websites/ReadComicOnline.ts
```diff
@@ -13,19 +14,36 @@ function ChapterExtractor(anchor: HTMLAnchorElement) {
...
 
-function MangInfoExtractor(element: HTMLDivElement) { //needed because titles are truncated in link
-    const anchor = element.querySelector<HTMLAnchorElement>('a');
-    const dom = new DOMParser().parseFromString(element.getAttribute('title'), 'text/html');
-    return {
-        id: anchor.pathname,
-        title: dom.querySelector('p.title').textContent.trim()
-    };
-}
+const pageScript = `
```
ronny1982, line 17: Why not using the script from [HakuNeko](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/KissComic.mjs#L55-L69)?

web/src/engine/websites/ReadComicOnline.ts
MikeZeDev: Should i put a reduced timeout for the promise then ?

web/src/engine/websites/ReadComicOnline.ts
MikeZeDev: They are sequentially done, if last id ok i assume previouses are ok but you are right.

web/src/engine/websites/ReadComicOnline.ts
MikeZeDev: Like i've explained in the PR message, with Hakuneko solution that a contribor did recently, we get a lot of blank.gif in haruneko. Dont ask me why 🤷 . The solution i've found was to increase delay. Like i've said, a fixed delay is not reliable.

I use setInterval to resolve the promise asap, but when "its done".

web/src/engine/websites/ReadComicOnline.ts
ronny1982: It is ore a re-arranging issue. Solution maybe to move the code into a finally block, or just place it after the catch block.

web/src/engine/websites/ReadComicOnline.ts
ronny1982: I was referring to consider a CSS query instead of regex-ing the image array
e.g.,
```javascript
const links = [ ...document.querySelectorAll('#divImage img') ].map(img => img.src);
if(links.length > 0 && !links.some(link => /blank.gif/i.test(link))) {
    resolve(images);
}
```

web/src/engine/websites/ReadComicOnline.ts
```diff
@@ -16,19 +16,18 @@ function ChapterExtractor(anchor: HTMLAnchorElement) {
...
-                    resolve (imagearray);
+                    resolve(links);
                 }
             } catch (error) {
+                clearInterval(interval);
+                reject(error);
+            } finally {
                 tries++;
                 if (tries > 10) {
                     clearInterval(interval);
```
ronny1982, line 33: The `error` variable is likely undefined. Need to create an appropriate Error.

web/src/engine/websites/ReadComicOnline.ts
```diff
@@ -31,7 +31,7 @@ const pageScript = `
...
-                    reject(error);
+                    reject('Unable to get pictures after more than 10 tries !');
```
ronny1982, line 34: Does this work?
I would expect to reject an `Error` type instead of a `string`, not sure how HaruNeko handles a thrown `string`.

ronny1982 (approved):

PR #708, opened 2024-07-27
MH  : move to templates folder

MikeZeDev:
because MH has no decorators


web/src/engine/websites/xianman123.ts
```diff
@@ -2,16 +2,11 @@ import { Tags } from '../Tags';
...
 
 const pagesScript = `
-    new Promise((resolve, reject) => {
-        try {
-            const pageLinks = picdata.map(pic => imgDomain != '' ? new URL(pic, imgDomain).href : pic);
-            resolve(pageLinks);
-        } catch (error) {
-            reject(error);
-        }
+    new Promise( resolve  => {
```
ronny1982, line 8: No need for curly block brackets when returning a single statement

ronny1982 (approved):

PR #714, opened 2024-07-29
bundle as linux tar

Robonau:
as the title says its some bundles for Linux tar

made them not portable because they will be more useful as not portable in regards to external builds
(like the AURs PKGBUILD)

the icon and desktop entry files should probably not be in the tar and instead in a separate linux-assets.tar.gz

i don't plan to maintain this, was just something i wanted to do cause someone asked for a linux release in the discord
though its almost a copy of the windows portable zip so it shouldn't be to bad


ronny1982:
### Tested Environment:
- Dell Inspiron 15
- Intel i3 11th Generation (4-Core)
- 8 GB Ram
- 256 GB SSD
- Kubuntu 24.04. (64-Bit)

----

### NW 64-Bit Build
⚠️ Crashed as soon as the UI was doing some more render intensive work e.g., MangaDex/Berserk chapter list.
[nw.log](https://github.com/user-attachments/files/16487933/nw.log)

### NW 32-Bit Build
⚠️ Could not be started on the 64-Bit system (probably because dependencies to i386 were missing).

### Electron 64-Bit Build
❌ Could not be started, probably because some files in the bundle needs certain permissions before getting bundled in a `tar.gz` archive. 
[electron.log](https://github.com/user-attachments/files/16487949/electron.log)

### Electron 32-Bit Build
⚠️ There was related archive after the bundle operation.

### Electron ARM Builds
ℹ️ Could not be tested due to lack of hardware with corresponding architecture.

ronny1982:
When deploying `tar.gz` archives it might be useful for the consumer to know what to do, so providing instructions may help e.g., as readme inside of the archive (a veteran would probably run `configure`, `make` and `make install` which will not work with this archive).

ronny1982:
The `hakunkeo.desktop` file is more like a template and not of much use as is. Maybe it would be better to not include this file at all?

ronny1982 (commented):
Seems the topic needs some more honing and testing on linux platforms before an NW or Electron canary bundle can be released to the public

Robonau:
yea, i have no clue what im doing tbh, i just wanted a build that worked on linux
and this worked for me

ronny1982:
Closed as not stable enough and abandoned by contributor

PR #720, opened 2024-07-31
Add Mikoroku

MikeZeDev:


web/src/engine/websites/Mikoroku.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const { feed: { entry } } = await FetchJSON<RSS>(new Request(new URL('/feeds/posts/default/-/Series?orderby=published&alt=json&max-results=999', this.URI)));
+        return entry.map(manga => {
+            const goodLink = manga.link.find(link => link.rel === 'alternate');
+            return new Manga(this, provider, new URL(goodLink.href).pathname, goodLink.title.trim());
```
ronny1982, line 43: Would it be possible to determine the real manga id instead of just a link which requires the injection of an additional script in `FetchChapters` to get the real manga id?

web/src/engine/websites/Mikoroku.ts
```diff
@@ -0,0 +1,56 @@
...
+        });
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const mangaid = await FetchWindowScript<string>(new Request(new URL(manga.Identifier, this.URI)), 'clwd.settings.cat');
+        const { feed: { entry } } = await FetchJSON<RSS>(new Request(new URL(`/feeds/posts/default/-/${mangaid}?orderby=published&alt=json&max-results=9999`, this.URI)));
+        return entry.map(entry => {
+            const goodLink = entry.link.find(link => link.rel === 'alternate');
+            return new Chapter(this, manga, new URL(goodLink.href, this.URI).pathname, goodLink.title.replace(manga.Title, '').trim());
+        }).filter(chap => chap.Identifier != manga.Identifier);
```
ronny1982, line 53: May use `chapter` instead of `chap`. There is no real benefit of removing the last 3 letters from the variable name at the cost of readability.

ronny1982, line 53: The chapter identifier is compared against the manga identifier, is this by intention?
It makes not much sense that there is a chapter with the same ID as the manga the chapter belongs to.

ronny1982 (approved):

web/src/engine/websites/Mikoroku.ts
```diff
@@ -0,0 +1,56 @@
...
+                href: string
+            }[]
+        }[]
+    }
+}
+
+@Common.MangaCSS(/^{origin}\/\d+\/\d+\/[^/]+\.html$/, 'header h1[itemprop="name"]')
+@Common.PagesSinglePageJS(pagesScript, 500)
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
```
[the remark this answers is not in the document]
MikeZeDev: I know that very well. We should make  BloggerManga  decorators. But they have little differences. Its planned tho.

web/src/engine/websites/Mikoroku.ts
MikeZeDev: From my experience with NextScan , manga id for this request is a real mess. You think its the slug, but sometimes its  the title (o the same website) , and dont let me talk about the non-consistant CaSeS problems. NextScans gave me headaches.

I did that because i want it to be reliable.

web/src/engine/websites/Mikoroku.ts
MikeZeDev: Yup. the feed always got, among chapters entry , an entry with the link to the manga itself. We have to filter it.

web/src/engine/websites/Mikoroku.ts
MikeZeDev: image

That is the title : `Er0t1c4l Wizard & Twelve Brides`.
slug is `erotical-wizard-twelve-brides`.

You may say : both of them works? Well, i'd found (on NextScans) thats kinda random

MikeZeDev: Wait for this website there is an alernative

PR #737, opened 2024-08-11
GourmetScans : fix getting pages

MikeZeDev:
Simulate activity to force scripts loading and execution.


web/src/engine/websites/GourmetScans.ts
```diff
@@ -5,25 +5,37 @@ import * as Madara from './decorators/WordPressMadara';
...
+                reject(error);
+            } finally {
+                tries++;
+                if (tries > 10) {
+                    clearInterval(interval);
+                    reject(new Error('Unable to get pictures after more than 10 tries !'));
+                }
+            }
+        }, 1000);
+       window.dispatchEvent(new KeyboardEvent('mousemove'));
```
ronny1982: Is `KeyboardEvent` used by intention instead of `MouseEvent`?

web/src/engine/websites/GourmetScans.ts
```diff
@@ -5,25 +5,37 @@ import * as Madara from './decorators/WordPressMadara';
...
+                        format: CryptoJSAesJson
+                    }).toString(CryptoJS.enc.Utf8));
+                    resolve(JSON.parse(imgdata));
+                }
+            } catch (error) {
+                clearInterval(interval);
+                reject(error);
+            } finally {
+                tries++;
+                if (tries > 10) {
```
ronny1982, line 24: Seems to me like the abort condition is intended to be 10 seconds instead of 10 attempts.
Either use `FetchWindowScript()` with timeout of 10 seconds, or use time as condition:
```javascript
const start = Date.now();
// ...
if(Date.now() - start > 10_000) {
    reject();
}
```

ronny1982 (approved):

web/src/engine/websites/GourmetScans.ts
MikeZeDev: Turns out PagesSinglePageJS doesnt use timeout parameter. 
Gotta use first solution then .

PR #739, opened 2024-08-11
ManyToon/ManhwaHentaiMe : fix websites

MikeZeDev:
Websites are non standard Madara

* Fix clipboard
* Fix manga list
* Fix chapters

### DIFF
--- a/web/src/engine/websites/ManhwaHentaiMe.ts
+++ b/web/src/engine/websites/ManhwaHentaiMe.ts
@@ -1,21 +1,70 @@
-import { Tags } from '../Tags';
+import { type Tag, Tags } from '../Tags';
 import icon from './ManhwaHentaiMe.webp';
-import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import { Chapter, DecoratableMangaScraper, Manga, type MangaPlugin } from '../providers/MangaPlugin';
 import * as Madara from './decorators/WordPressMadara';
 import * as Common from './decorators/Common';
+import { FetchCSS, FetchHTML } from '../platform/FetchProvider';
 
-@Madara.MangaCSS(/^{origin}\/webtoon\/[^/]+\/$/, 'div.post-title h1')
-@Madara.MangasMultiPageAJAX()
-@Madara.ChaptersSinglePageAJAXv2()
+type MangaID = {
+     post: string;
+     slug: string;
+}
+
+function MangaInfoExtractor(anchor: HTMLAnchorElement) {
+    const container = anchor.closest<HTMLElement>('div.page-item-detail, div.manga');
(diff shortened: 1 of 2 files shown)


web/src/engine/websites/ManhwaHentaiMe.ts
```diff
@@ -1,21 +1,70 @@
...
+        return new RegExpSafe(`^${this.URI.origin}/webtoon/[^/]+\/$`).test(url);
+    }
+
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const uri = new URL(url);
+        const data = await FetchHTML(new Request(uri));
+        const post = data.querySelector<HTMLElement>('div#star[data-id]')?.dataset?.id;
+        const slug = uri.pathname;
+        const element = data.querySelector<HTMLElement>('div.post-title h1');
+        const title = Common.AnchorInfoExtractor(false, 'span').call(this, element).title.replace(/Webtoon Manhwa Hentai$/, '').trim();
```
ronny1982, line 45: May extract the replacement since it is used in other function as well.

web/src/engine/websites/ManhwaHentaiMe.ts
```diff
@@ -1,21 +1,70 @@
...
+        return new Manga(this, provider, JSON.stringify({ post, slug }), title);
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
```
ronny1982, line 49: Is this something specific to this website, or is this something generic that may be adapted by other Madara websites as well?
In the latter case it would make sense to implement this method as decorator.

web/src/engine/websites/ManhwaHentaiMe.ts
```diff
@@ -1,21 +1,70 @@
...
+            }).toString(),
+            headers: {
+                'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8',
+                Referer: this.URI.href
+            }
+        });
+
+        const data = await FetchCSS(request, 'ul li.wp-manga-chapter > a');
+        return data.map(element => {
+            const { id, title } = Common.AnchorInfoExtractor(false, 'span').call(this, element);
```
ronny1982, line 65: Same extractor is created and used in other function as well, consider to extract for re-usage.

ronny1982 (approved):

web/src/engine/websites/ManhwaHentaiMe.ts
MikeZeDev: I dont think 2 websites is enough to make a specific decorator. 
On the other hand

* we can always add the postid element finding in MangaCSS (since we already have a couple of them)
* A JS decorator to get chapters from Madara Manga IDentifier  ( post + slug) can always be handy. We never know who can block admin-ajax with Cloudflare

PR #741, opened 2024-08-12
RawLazy : fix getting pages

MikeZeDev:
Step1 : get a link
Step2: get the iframe link in previous link
Step3: Profit

### DIFF
--- a/web/src/engine/websites/RawLazy.ts
+++ b/web/src/engine/websites/RawLazy.ts
@@ -20,13 +20,36 @@ type APIResult = {
     mes: string;
 }
 
+const IframeScript = `
+    new Promise( (resolve, reject) => {
+        const start = Date.now();
+        const interval = setInterval(function () {
+            try {
+                if (document.querySelector('#idIframe')) {
+                    clearInterval(interval);
+                    resolve(document.querySelector('#idIframe').src);
+                }
+            } catch (error) {
+                clearInterval(interval);
+                reject(error);
+            } finally {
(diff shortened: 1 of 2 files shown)


web/src/engine/websites/RawLazy.ts
```diff
@@ -70,25 +93,32 @@ export default class extends DecoratableMangaScraper {
...
+                            resolve(document.querySelector('${selector}').href);
+                        }
+                    } catch (error) {
                         clearInterval(interval);
-                        resolve(document.querySelector('${selector}').href));
+                        reject(error);
+                    } finally {
+                        if(Date.now() - start > 10_000) {
+                            clearInterval(interval);
+                            reject(new Error('Unable to get pictures after more than 10 seconds !'));
```
ronny1982, line 118: This is probably shown to the user on failed download.
Use a more appropriate and localised message e.g., `FetchProvider_FetchWindow_TimeoutError`.

MikeZeDev: Yup. How i am supposed to use localised message in injected promises? 😔

ronny1982: Consider a different approach which will provide an appropriate error message
```javascript
new Promise((resolve, reject) => {
    const interval = setInterval(() => {
        console.log('Do some stuff ...');
    }, 1000);

    AbortSignal.timeout(10_000).onabort = function({ target: { reason } }) {
        clearInterval(interval);
        reject(reason);
    }
});
```

PR #748, opened 2024-08-16
Port Qiman5

MikeZeDev:


web/src/engine/websites/Qiman5.ts
```diff
@@ -0,0 +1,39 @@
...
+import { FetchJSON } from './../platform/FetchProvider';
+
+type APIChapters = {
+    data: {
+        name: string,
+        link : string
+    }[]
+}
+
+@Common.MangaCSS(/^{origin}\/[^.]+\.html$/, 'div.info h1.name_mh')
```
ronny1982, line 14: The part `[^.]` will also match slashes, is this by intention?

web/src/engine/websites/Qiman5.ts
```diff
@@ -0,0 +1,39 @@
...
+    public constructor() {
+        super('qiman5', `奇漫屋 (Qiman5)`, 'https://qiman2.com', Tags.Language.Chinese, Tags.Media.Manhua, Tags.Media.Manga, Tags.Media.Manhwa, Tags.Source.Aggregator);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const mangaId = manga.Identifier.match(/\/(\d+)\.html/)[1];
```
ronny1982, line 29: This is a code smell as the manga identifier seems to store bloat which is not necessary to identify the manga.

web/src/engine/websites/Qiman5.ts
```diff
@@ -0,0 +1,39 @@
...
+
+type APIChapters = {
+    data: {
+        name: string,
+        link : string
+    }[]
+}
+
+@Common.MangaCSS(/^{origin}\/[^.]+\.html$/, 'div.info h1.name_mh')
+@Common.MangasMultiPageCSS('/category/tags/6/page/{page}', 'div.item p.title a', 1, 1, 0, Common.AnchorInfoExtractor(true))
```
ronny1982, line 15: Use an extractor that provides only the number required to identify the manga.

web/src/engine/websites/Qiman5.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const mangaId = manga.Identifier.match(/\/(\d+)\.html/)[1];
+        const { data } = await FetchJSON<APIChapters>(new Request(new URL(`/index.php/api/comic/chapter?mid=${mangaId}`, this.URI)));
+        return data.map(chapter => new Chapter(this, manga, chapter.link, chapter.name.trim())).distinct();
```
ronny1982, line 31: I would expect the REST endpoint does not provide duplicates, is `distinct` really required?

web/src/engine/websites/Qiman5.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const pages = await Common.FetchPagesSinglePageJS.call(this, chapter, `decodeURIComponent(authCrypt.decode(DATA, 'qiman2')).split(',');`, 500);
```
ronny1982, line 35: Consider use `FetchWindowScript` directly since the decorator seems not suitable (returned pages needs to be converted).

ronny1982 (approved):

web/src/engine/websites/Qiman5.ts
MikeZeDev: Ah, missed this one, you are right

web/src/engine/websites/Qiman5.ts
MikeZeDev: But then i would have to write a custom FetchManga as MangaCSS doesnt allow to choose our own manga ID.

web/src/engine/websites/Qiman5.ts
MikeZeDev: I could use a custom MangaInfoExtractor. But MangaCSS use a Label Extractor that doesnt allow us to make a customa Manga ID. So i would need a custom FetchManga.

I'd rather write less code , therefore just making a custom FetchChapters is the most efficient way to go.

web/src/engine/websites/Qiman5.ts
MikeZeDev: Ive seen duplicates, yes as strange as it is. I assumed the same but i was wrong. Perhaps i found the only case of the whole website idk. On Yuan zun i guess?

PR #752, opened 2024-08-20
Mangatown: handle both chapters types

MikeZeDev:
We have chapters will all images in one html page, and chapters with one image per html page and a select element listing html pages.

But here is the magic : A slightly modified DM5 script works on both. We win time because we dont fetch all those html pages anymore 💪


web/src/engine/websites/MangaTown.ts
```diff
@@ -19,7 +40,7 @@ export default class extends DecoratableMangaScraper {
     }
 
     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
-        const data = await FetchCSS<HTMLOptionElement>(new Request(new URL(chapter.Identifier, this.URI)), 'div.manga_read_footer div.page_select select option');
-        return data.filter(option => !/featured\.html$/.test(option.value)).map(option => new Page(this, chapter, new URL(option.value, this.URI), { Referer: 'mangahere.com' }));
+        const pages = await Common.FetchPagesSinglePageJS.call(this, chapter, pageScript);
+        return pages.map(page => new Page(this, chapter, page.Link, { Referer: 'https://mangahere.com' }));
```
ronny1982, line 44: Avoid intermediate mapping from `Page[]` to `Page[]` just to change a single property, instead get the pages directly.
```javascript
const data = await FetchWindowScript<string[]>(new Request(new URL(chapter.Identifier, this.URI)), pageScript);
return data.map(link => new Page(this, chapter, new URL(link), { Referer: 'https://mangahere.com' }));
```

web/src/engine/websites/MangaTown.ts
```diff
@@ -1,13 +1,34 @@
...
+            const pagecount = window.total_pages;
+            const images = await Promise.all([...new Array(pagecount).keys()].map(async p => {
+                eval(await $.ajax({
+                    url: 'chapterfun.ashx',
+                    data: {
+                        cid: window.chapter_id,
+                        page: p + 1,
+                    }
+                }));
+                return d[0];
```
ronny1982, line 19: I assume `d[0]` is a global variable which is update within `eval`, therefore all `eval`s should be executed synchronous instead of parallel to avoid race conditions and unexpected `d[0]` assignments.

web/src/engine/websites/MangaTown.ts
```diff
@@ -1,13 +1,34 @@
...
-import { FetchCSS } from '../platform/FetchProvider';
 import * as Common from './decorators/Common';
 
+//Modded DM5 script
+const pageScript = `
+    new Promise(async (resolve, reject) => {
+        try {
+            const pagecount = window.total_pages;
+            const images = await Promise.all([...new Array(pagecount).keys()].map(async p => {
+                eval(await $.ajax({
```
ronny1982, line 12: May use fetch API instead.

ronny1982 (changes requested):

web/src/engine/websites/MangaTown.ts
ronny1982: May also check if the call to `eval` return the value that is stored within `d[0]`, then `d[0]` can be avoided at all which automatically solves the issue.

MikeZeDev: I see 👌 . Then perhaps the Dm5 original script should be revised as well according to your specifications.

web/src/engine/websites/MangaTown.ts
```diff
@@ -2,23 +2,22 @@ import { Tags } from '../Tags';
...
-                        page: p + 1,
-                    }
-                }));
-                return d[0];
-            }));
-            resolve(images.map(link => new URL(link, window.location.href).href));
+            const pages = [];
+            for (let page = 1; page <= pagecount; page++) {
+                const response = await fetch('chapterfun.ashx?cid='+ window.chapter_id + '&page=' + page);
+                const script = await response.text();
```
ronny1982, line 15: May avoid intermediate variable `script` and evaluate response text directly.

web/src/engine/websites/MangaTown.ts
```diff
@@ -2,23 +2,22 @@ import { Tags } from '../Tags';
...
+                eval(script);
+                pages.push(d[0]);
```
ronny1982, line 17: May compose FQDN here instead of mapping later `new URL(d[0], window.location.href).href)`

web/src/engine/websites/MangaTown.ts
```diff
@@ -19,7 +39,7 @@ export default class extends DecoratableMangaScraper {
     }
 
     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
```
ronny1982, line 41: This method should avoid making dozens of requests for each page (e.g., responsive usability, avoid timeouts).
Consider using the HTML pages itself as page links (e.g., https://www.mangatown.com/manga/goblin_slayer/c090/2.html) and then doing this JS eval magic to get the image link in the `FetchImage` method.

ronny1982 (approved):

web/src/engine/websites/MangaTown.ts
MikeZeDev: > making dozens of requests for each page? 

I fail to see how we are "making dozens of requests for each page".  `FetchPages` does one request per page (to the endpoint) + 1 initial request. FetchImage only fetch the image.

What is true is that FetchPages is making too many requests.

Ill change it so Fetchpages will just return Pages with needed Parameters { chapterId , pageNumber} 
FetchImage will use the endpoint and eval magic to get real picture url and download it.

PR #766, opened 2024-09-16
Add RuManhua

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7320


web/src/engine/websites/RuManhua.ts
```diff
@@ -0,0 +1,92 @@
...
+    bookName: string
+}
+
+type APIChapter = {
+    chapterid: string,
+    chaptername: string
+}
+
+const pagesScript = `
+    new Promise (resolve => resolve( [... document.querySelectorAll('div.chapter-img-box img')].map(image=> image.dataset.src || image.src)));
```
ronny1982, line 22: This should work without wrapping the result in a promise, since it is a single command

web/src/engine/websites/RuManhua.ts
```diff
@@ -0,0 +1,92 @@
...
+        return icon;
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return new RegExpSafe(`^${this.URI.origin}/[^/]+/$`).test(url);
+    }
+
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const title = (await FetchCSS(new Request(url), 'div.book-name h1.name')).shift().textContent.trim();
+        const id = url.match(/\/([^/]+)\/$/)[1];
```
ronny1982, line 43: No need for regex `url.split('/').at(1)`

web/src/engine/websites/RuManhua.ts
```diff
@@ -0,0 +1,92 @@
...
+        }
+        return mangaList;
+    }
+
+    private async GetMangasFromPage(page: number, provider: MangaPlugin): Promise<Manga[]> {
+        const { data } = await FetchJSON<APIResponse<APIManga[]>>(new Request(new URL('/data/sort', this.URI), {
+            method: 'POST',
+            body: `s=1&p=${page}`,
+            headers: {
+                'Content-type': 'application/x-www-form-urlencoded; charset=UTF-8',
```
ronny1982, line 61: Mixed casing, either use lower case or upper (snake) case for headers

web/src/engine/websites/RuManhua.ts
```diff
@@ -0,0 +1,92 @@
+import { Tags } from '../Tags';
+import icon from './RuManhua.webp';
+import { Chapter, DecoratableMangaScraper, Manga, type MangaPlugin } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import { FetchCSS, FetchJSON } from '../platform/FetchProvider';
+
+type APIResponse<T> = {
+    data : T | string
```
ronny1982, line 8: Is the union with `string` required?
I didn't see any business logic that processes the `data` field as string.

web/src/engine/websites/RuManhua.ts
```diff
@@ -0,0 +1,92 @@
...
+                Origin: this.URI.origin,
+                Referer: this.URI.href,
+                'X-Requested-With': 'XMLHttpRequest'
+            }
+        }));
+        return data instanceof Array ? (data as APIManga[]).map(item => new Manga(this, provider, item.id, item.bookName)): [];
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        //fetch first batch of chapters
```
ronny1982, line 71: Remove redundant comment, this is quite obvious from the code itself

web/src/engine/websites/RuManhua.ts
```diff
@@ -0,0 +1,92 @@
...
+        const chapterList = await Common.FetchChaptersSinglePageCSS.call(this, manga, 'div.chaplist-box ul li a');
+
+        //fetch "MoreChapters"
```
ronny1982, line 74: Remove redundant comment, this is quite obvious from the code itself

web/src/engine/websites/RuManhua.ts
```diff
@@ -0,0 +1,92 @@
...
+        const { data } = await FetchJSON<APIResponse<APIChapter[]>>(new Request(new URL('/morechapter', this.URI), {
+            method: 'POST',
+            body: `id=${manga.Identifier}`,
+            headers: {
+                'Content-type': 'application/x-www-form-urlencoded; charset=UTF-8',
```
ronny1982, line 79: Avoid mixing upper/lower case in header name

ronny1982 (approved):

web/src/engine/websites/RuManhua.ts
MikeZeDev: there is no "undefined or null" here. In case the api response is empty, data is an empty string. Otherwise its type is T.

Perhaps i should replace string with "" ?

PR #770, opened 2024-09-18
Mangazure : fix getting images

MikeZeDev:


web/src/engine/websites/Mangazure.ts
```diff
@@ -14,10 +14,11 @@ const pageScript = `
 
         function getImageUrl(path) {
             let result = '';
+            const $ImgHost = 'https://lh5.googleusercontent.com/';
```
ronny1982, line 17: Avoid using special characters for self-declared variable names

ronny1982 (approved):

PR #776, opened 2024-09-20
Comick : use fallback CDN for pictures

MikeZeDev:
Newest chapters dont necessarily have pictures on s3 server.

### DIFF
--- a/web/src/engine/websites/ComicK.ts
+++ b/web/src/engine/websites/ComicK.ts
@@ -1,7 +1,6 @@
 import { Tags } from '../Tags';
 import icon from './ComicK.webp';
 import { Chapter, DecoratableMangaScraper, Manga, Page, type MangaPlugin } from '../providers/MangaPlugin';
-//import * as Common from './decorators/Common';
 import { Fetch, FetchJSON, FetchWindowScript } from '../platform/FetchProvider';
 import type { Priority } from '../taskpool/DeferredTask';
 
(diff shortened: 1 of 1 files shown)


web/src/engine/websites/ComicK.ts
```diff
@@ -136,4 +136,13 @@ export default class extends DecoratableMangaScraper {
         const { chapter: { md_images } } = await FetchJSON<APISingleChapter>(new Request(new URL(`/chapter/${chapter.Identifier}`, this.apiUrl)));
         return md_images.map(image => new Page(this, chapter, new URL(image.b2key, `https://s3.comick.ink/comick/`), { Referer: this.URI.href }));
     }
+
+    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        const blob = await Common.FetchImageAjax.call(this, page, priority, signal, true);
```
ronny1982, line 141: Each `FetchImageAjax` call will queue a job in the `imageTaskPool`, which may lead to long delays until the final image is received in case between both calls the queue is filled with other image download tasks (which is very likely since the `FetchImage` methods are all fired at once when downloading the pages).

Rule of thumb: Only one task should be queued in the `imageTaskPool` within the `FetchImage` method.

ronny1982 (changes requested):

web/src/engine/websites/ComicK.ts
MikeZeDev: Should i do like Alphapolis then ? Manual fetching wrapped in a job ? 

```ts
    public override async FetchImage(page: Page, priority: Priority, signal: AbortSignal): Promise<Blob> {
        return this.imageTaskPool.Add(async () => {
            try {
                const request = new Request(page.Link, {
                    signal: signal,
                    headers: {
                        Referer: this.URI.href
                    }
                });
                const response = await Fetch(request);
                return response.blob();
            } catch (error) {

                if (page.Parameters.fallbackURL) {
                    const request = new Request(page.Parameters.fallbackURL as string, {
                        signal: signal,
                        headers: {
                            Referer: this.URI.href
                        }
                    });
                    const response = await Fetch(request);
                    return response.blob();
                } else Promise.reject(error);
            }
        }, priority, signal);
    }

```

ronny1982 (approved):

PR #794, opened 2024-09-30
Yuramanga : no more madara

MikeZeDev:
Its bloggerManga. Gotta see in the future if we can make decorators from those websites.


web/src/engine/websites/YuraManga.ts
```diff
@@ -1,13 +1,42 @@
...
 
-@Madara.MangaCSS(/^{origin}\/manga\/[^/]+\/$/, 'meta[property="og:title"]:not([content*="YuraManga"])')
-@Madara.MangasMultiPageAJAX()
-@Madara.ChaptersSinglePageAJAXv2()
-@Madara.PagesSinglePageCSS()
+type RSS = {
+    feed: {
+        entry: {
+            title: {
+                $t: string
```
ronny1982, line 11: Remove unused fields from type

Issue #801, opened 2024-10-04
Corocoro: Website change


sakuhanachan:
Shogakukan renewal Corocoro's online manga site, manga list and reader has bean change from CoreView.

site: https://www.corocoro.jp/
list: https://www.corocoro.jp/rensai
manga page: https://www.corocoro.jp/title/14
manga viewer page: https://www.corocoro.jp/chapter/617/viewer

note: this site use Shogakukan ID for purchase some episode

MikeZeDev:
NextJS chunks craaaaaaaaaap ;-;

PR #821, opened 2024-10-11
Corocoro : fix website

MikeZeDev:
Closes https://github.com/manga-download/haruneko/issues/801


web/src/engine/websites/CoroCoro.ts
```diff
@@ -1,21 +1,59 @@
...
+import { Exception } from '../Error';
+import { WebsiteResourceKey as R } from '../../i18n/ILocale';
 
-function MangaExtractor(element: HTMLAnchorElement) {
-    return {
-        id: element.pathname,
-        title: element.querySelector('div.p-article-image img').getAttribute('alt').trim()
-    };
+type JSONMangas = {
+    weekdays: {
```
ronny1982, line 11: May use keyed type such as Record
```typescript
type WeekDays = 'mon' | 'tue' | 'wed';
type Mangas = Record<WeekDays, Manga[]>;
```

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
     public override get Icon() {
         return icon;
     }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL('/rensai', this.URI)), 'script');
+        const json = this.FindAndParseJSON(scripts, 'totalChapterLikes');
+        const { weekdays: { mon, tue, wed, thu, fri, sat, sun } } = this.LocateObjectWithNamedProperty<JSONMangas>(json, 'weekdays');
+        return [mon, tue, wed, thu, fri, sat, sun].reduce((accumulator: Manga[], day) => {
```
ronny1982, line 71: No need to write each key, consider using `Object.values(weekdays).reduce` instead

web/src/engine/websites/CoroCoro.ts
ronny1982: On second thought, do the weekdays really matter?
May just use `type Mangas = Record<string, Manga[]>;`

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+            const mangas = day.map(manga => new Manga(this, provider, `/title/${manga.id}`, manga.name));
+            accumulator.push(...mangas);
+            return accumulator;
+        }, []);
+
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL(`${manga.Identifier}`, this.URI)), 'script');
```
ronny1982, line 80: Consider restricting the query for scripts e.g., `head script`

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+        if (!cryptoParams?.key) return blob;
+        const encrypted = await blob.arrayBuffer();
+        const cipher = { name: 'AES-CBC', iv: Buffer.from(cryptoParams.iv, 'hex') };
+        const cryptoKey = await crypto.subtle.importKey('raw', Buffer.from(cryptoParams.key, 'hex'), cipher, false, ['decrypt']);
+        const decrypted = await crypto.subtle.decrypt(cipher, cryptoKey, encrypted);
+        return Common.GetTypedData(decrypted);
+    }
+
+    private FindAndParseJSON(scripts: HTMLScriptElement[], textToFind: string): JSONObject {
+        const script = scripts.find(script => script.innerText.indexOf(textToFind) > 0)?.innerText;
```
ronny1982, line 112: Use [`.text`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement/text) instead of `.innerText` for scripts

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+        const cryptoParams = page.Parameters;
+        if (!cryptoParams?.key) return blob;
+        const encrypted = await blob.arrayBuffer();
+        const cipher = { name: 'AES-CBC', iv: Buffer.from(cryptoParams.iv, 'hex') };
+        const cryptoKey = await crypto.subtle.importKey('raw', Buffer.from(cryptoParams.key, 'hex'), cipher, false, ['decrypt']);
+        const decrypted = await crypto.subtle.decrypt(cipher, cryptoKey, encrypted);
+        return Common.GetTypedData(decrypted);
+    }
+
+    private FindAndParseJSON(scripts: HTMLScriptElement[], textToFind: string): JSONObject {
```
ronny1982, line 111: Consider the needle to be RegExp instead of string for more sophisticated filtering

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+        const script = scripts.find(script => script.innerText.indexOf(textToFind) > 0)?.innerText;
+        if (!script) return undefined;
+        const json = JSON.parse(script.substring(22, script.length - 2));
```
ronny1982, line 114: Minor changes in the website (spaces, minification, ...) may change the offset from `22` to something else.
Consider other ways to determine the start/end indices.

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+            const mangas = day.map(manga => new Manga(this, provider, `/title/${manga.id}`, manga.name));
+            accumulator.push(...mangas);
+            return accumulator;
+        }, []);
+
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL(`${manga.Identifier}`, this.URI)), 'script');
+        const json = this.FindAndParseJSON(scripts, 'omittedMiddleChapters');
```
ronny1982, line 81: The `json` variable may be undefined here, is this intentional?

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+            const chapters = category.map(chapter => new Chapter(this, manga, `/chapter/${chapter.id}`, chapter.title));
+            accumulator.push(...chapters);
+            return accumulator;
+        }, []);
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page<CryptoParams>[]> {
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL(`${chapter.Identifier}`, this.URI)), 'script');
+        const json = this.FindAndParseJSON(scripts, 'viewerSection');
+        if (!json) {
```
ronny1982, line 93: This method handles undefined `json` but `FetchChapters` does not. Is this by intention?

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+            return accumulator;
+        }, []);
+
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL(`${manga.Identifier}`, this.URI)), 'script');
+        const json = this.FindAndParseJSON(scripts, 'omittedMiddleChapters');
+        const { chapters: { earlyChapters, omittedMiddleChapters, latestChapters } } = this.LocateObjectWithNamedProperty<JSONChapters>(json, 'chapters');
+        return [earlyChapters, omittedMiddleChapters, latestChapters].reduce((accumulator: Chapter[], category) => {
```
ronny1982, line 83: Are the key names relevant? If not just use generic string keys as recommended for the weekday object.

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
     public override get Icon() {
         return icon;
     }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL('/rensai', this.URI)), 'script');
+        const json = this.FindAndParseJSON(scripts, 'totalChapterLikes');
```
ronny1982, line 69: The `FindAndParseJSON` is always called in combination with `LocateObjectWithNamedProperty`. Consider combining both methods.

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+            throw new Exception(R.Plugin_Common_Chapter_UnavailableError);
+        }
+        const { viewerSection: { pages } } = this.LocateObjectWithNamedProperty<JSONPages>(json, 'viewerSection');
+        return pages.map(page => new Page<CryptoParams>(this, chapter, new URL(page.src), page.crypto));
+    }
+
+    public override async FetchImage(page: Page<CryptoParams>, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        const blob = await Common.FetchImageAjax.call(this, page, priority, signal, true);
+        const cryptoParams = page.Parameters;
+        if (!cryptoParams?.key) return blob;
```
ronny1982, line 103: Per type definition this field cannot be undefined.
Either update the type definition and mark it as nullable, or remove this condition.

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+    }
+
+    private FindAndParseJSON(scripts: HTMLScriptElement[], textToFind: string): JSONObject {
+        const script = scripts.find(script => script.innerText.indexOf(textToFind) > 0)?.innerText;
+        if (!script) return undefined;
+        const json = JSON.parse(script.substring(22, script.length - 2));
+        return JSON.parse(json.substring(json.indexOf(':') + 1));
+    }
+
+    private LocateObjectWithNamedProperty<T>(obj, keyname: string): T {
```
ronny1982, line 118: Use more appropriate variable name. When lacking creativity ask [copilot](https://copilot.microsoft.com/) for an appropriate suggestion by describing the usage.
image

web/src/engine/websites/CoroCoro.ts
```diff
@@ -25,4 +63,69 @@ export default class extends DecoratableMangaScraper {
...
+        if (!obj) return null;
+        if (obj[keyname]) {
+            return obj as T;
+        }
+        let result = null;
+        for (let i in obj) {
+            if (typeof obj[i] === 'object')
+                result = result ?? this.LocateObjectWithNamedProperty(obj[i], keyname);
```
ronny1982, line 126: Improve abort condition to short circuit the loop and recursion.
As soon as a result is assigned, then there is no reason to process the loop or any other recursive calls as they won't have any effect, instead return the result
```javascript
result = { x: 0 };
result = result ?? GetStuff(); // return value from GetStuff will never be assigned
```

ronny1982 (approved):

web/src/engine/websites/CoroCoro_e2e.ts
```diff
@@ -7,18 +7,18 @@ const config = {
...
-        title: '[第1話]'
+        id: '/chapter/4934',
+        title: '第1話'
     },
     entry: {
         index: 0,
-        size: 2_314_013,
-        type: 'image/png'
+        size: 169_646,
+        type: 'image/webp'
```
ronny1982, line 21: Should have test cases for encrypted and decrypted image scenario

web/src/engine/websites/CoroCoro.ts
MikeZeDev: Yes they are because its how chapters are sorted.. 
Website shows them in that order : earlyChapters , omittedMiddleChapters (hidden, its the reason why we do it like that) then latestChapters. If for some reason we omit the key and take them "as they are in the JSON" and for some reason the JSON put omittedMiddleChapters  before earlyChapters  order will be screwed.

web/src/engine/websites/CoroCoro.ts
MikeZeDev: The unique string used to find the script is not necessarily the variable property we are seeking, but i'll see what i can do.

web/src/engine/websites/CoroCoro.ts
MikeZeDev: Yeah i thought about breaking from the loop in case result was defined. I deemed it fast enough tho. Okay.

web/src/engine/websites/CoroCoro.ts
MikeZeDev: 22 is calculated from self.push(_next_f')"length,  from scripts. This is standard among all nextjs websites i have seen. I tested this script on all of them.

web/src/engine/websites/CoroCoro.ts
MikeZeDev: variable **value** may be undefined, or null, or whatever but the variable itself is always present afaik.

PR #833, opened 2024-10-19
Add ManhwaWeb

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7475


web/src/engine/websites/ManhwaWeb.ts
```diff
@@ -0,0 +1,80 @@
...
+        for (let page = 1, run = true; run; page++) {
+            const { mangas, next } = await this.GetMangasFromPage(page, provider);
+            mangaList.push(...mangas);
+            run = next;
+        }
+        return mangaList;
+    }
+
+    private async GetMangasFromPage(page: number, provider: MangaPlugin): Promise<MangaResult> {
+        const { data, next } = await FetchJSON<APIMangas>(new Request(new URL(`/manhwa/library?estado=&tipo=&erotico=&demografia=&page=${page}`, this.apiUrl)));
```
ronny1982, line 65: Can't the query parameters `estado=&tipo=&erotico=&demografia=` without an explicit value be removed from the URL?

ronny1982, line 65: Is `next` necessary, wouldn't it just return an empty list on the next request which can be used as abort condition like in all the other multi-paged websites?

ronny1982 (approved):

web/src/engine/websites/ManhwaWeb.ts
MikeZeDev: Nope. There was way more. Removing one of those i kept break the request.

MikeZeDev: Website give me a next, i use the next but yeah, sure.

Issue #797, opened 2024-10-01
jsPDF doesnt recognize some JPEG files


MikeZeDev:
Case: https://chapmanganato.to/manga-sr995652. Chapter 33 to 40

JPG file starts with header 
![image](https://github.com/user-attachments/assets/1c7681f3-d02a-47d9-99b5-88a51325e938)

When adding image as array buffer jsPDF checks for headers to determine image formats and this one is not listed under JPEG so image its not recognized.

![image](https://github.com/user-attachments/assets/ad461df2-4abe-47da-8670-5277bd219b14)

Proposed workaround : 
* use the addimage syntax that allows us to provide a fallback format perhaps

MikeZeDev:
if you use JPEG as fallback, generated pdf is garbage anyway

![image](https://github.com/user-attachments/assets/e1d96bbe-2a34-49fc-b04f-9d67fbed078d)

MikeZeDev:
* Possible workaround : check if jpeg file pass JSpdf checks and convert it ourselves before adding it to pdf

PR #856, opened 2024-11-02
Replace jsPDF with PDFKit

ronny1982:
Seems to have better support for a wider range of different JPEG flavors (#797)

## Bundle with jsPDF

HaruNeko - jsPDF

## Bundle with PDFKit

HaruNeko - PDFKit

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -6,8 +6,8 @@
         "npm": ">=10.2.4"
     },
     "dependencies": {
-        "jspdf": "^2.5.2",
         "jszip": "^3.10.1",
+        "pdfkit": "github:manga-download/pdfkit#v0.15.1",
         "protobufjs": "^7.4.0"
     },
     "devDependencies": {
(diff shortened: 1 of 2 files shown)


MikeZeDev:
I am fine with it but why not just making a PR in original pdfkit repo to add ES6 support? It seems maintained , last release was 4 days ago. Why hosting it on our own github? 

https://github.com/foliojs/pdfkit

ronny1982:
> I am fine with it but why not just making a PR in original pdfkit repo to add ES6 support? It seems maintained , last release was 4 days ago. Why hosting it on our own github?
> 
> https://github.com/foliojs/pdfkit

Its just a workaround to make the standalone build consumable via package dependency instead of just copying it into HaruNeko repo.

I don't have time to setup a pipeline to create an ES import friendly browser version of PDFKit.
Furthermore the docs stated that you should use browserify or the standalone build, so even when providing such a solution i doubt a PR would be accepted.

PR #866, opened 2024-11-09
fix the u_ error in prod

Robonau:
it gets minified in prod so you cant use this.constructor.name;

if you dont want to have to do this then you could also just add 
```
esbuild: {
    keepNames: true
}
```
to the vite.config.ts in web instead

#867 does that

### DIFF
--- a/web/src/engine/Error.ts
+++ b/web/src/engine/Error.ts
@@ -14,7 +14,11 @@ export class InternalError extends Error {
 /**
  * An error indicating that a certain programming logic is intentionally missing.
  */
-export /* sealed */ class NotImplementedError extends InternalError {}
+export /* sealed */ class NotImplementedError extends InternalError {
+    public override get name(): string {
+        return "NotImplementedError";
+    }
+}
 
 /**
  * An exception should be raised when a problem appears during runtime and should be properly handled at some point.


Sheepux:
Doesn't this line already do it ?
https://github.com/manga-download/haruneko/blob/master/web/src/engine/Error.ts#L10
```
export class InternalError extends Error {
    public override get name(): string {
        return this.constructor.name;
    }
}
```

ronny1982:
These are just [marker interfaces](https://en.wikipedia.org/wiki/Marker_interface_pattern), they are not expected to be used or shown in production.

image

MikeZeDev:
Should "NotImplementedError" be an "InternalError" then?  Should the user knows or not that something is not coded yet? 
Even if we display the error they will still open issues anyway. I am conflicted about it.

ronny1982:
> Should "NotImplementedError" be an "InternalError" then? Should the user knows or not that something is not coded yet? Even if we display the a proper error message like i would like, they will still open issues anyway. I am conflicted about it.

The `NotImplementedError` is already an `InternalError` and should ideally not appear anywhere within the release candidate(s). If it does the product is not yet ready for release.

The frontend would be responsible to show `Exception`s (which provides a localized user friendly message) and handle the representation all other errors (it is up to the frontend to decide).

ronny1982:
Closed as covered by #867

Sheepux:
I'm personally expecting the error to be the one handling the localized message so that i can catch(error) and display error.message
If it's the frontend responsiblity does it mean that I need to try catch every possible error message ?
When I use API that returns an error message that's the one you usually display.

ronny1982:
> I'm personally expecting the error to be the one handling the localized message so that i can catch(error) and display error.message If it's the frontend responsiblity does it mean that I need to try catch every possible error message ? When I use API that returns an error message that's the one you usually display.

There is nothing wrong with just showing the error message (if there is one). As stated above, the frontend decides what to do.

The engine promises that all errors of type `Exception` have a localized message. Any other errors may come from programming mistakes, 3rd party libraries, the system or somewhere else where the HakuNeko engine has no control.

This gives the frontend the possibility for a more sophisticated representation of errors due to distinguishing between user-friendly `Exception`s and anything else. The concept to distinguish between [exceptions and errors](https://dart.dev/language/error-handling) was adopted from dart.

PR #867, opened 2024-11-09
u_ error fix using keepNames

Robonau:
this or #866
either works

this will make the bundle larger. but it means you don't gotta think about name whenever you extend the InternalError class

### DIFF
--- a/web/vite.config.ts
+++ b/web/vite.config.ts
@@ -40,5 +40,8 @@ export default defineConfig({
         include: ['carbon-components-svelte'],
         // carbon-icons-svelte is huge and takes 12s to prebundle, better use deep imports for the icons you need
         exclude: ['carbon-icons-svelte']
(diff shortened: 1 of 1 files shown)


Sheepux:
see comment in #866
The Issue is fixed by https://github.com/manga-download/haruneko/commit/04fe6033155c8c8d743277fd0090abe4b1d42645

ronny1982:
Keeping the `name` property would only solve half of the issue, as the stack trace is still filled with minified references. It would be either to disable minification completely to get readable errors, or leave it as is and live with unreadable errors in production.

It should be noted that users are expected to see localized exceptions instead of errors.

ronny1982:
Closed, will keep as is for now.
May re-consider to disable minification completely in case proper error handling still seems to be an issue in the future.

PR #877, opened 2024-11-14
Mangalib : rewrite website

MikeZeDev:
Prelude to future Libgroup template (we also have HentaiLib requested)

https://github.com/manga-download/hakuneko/issues/7545


web/src/engine/websites/MangaLib.ts
```diff
@@ -1,46 +1,125 @@
...
-        super('mangalib', 'MangaLib', 'https://mangalib.org', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Language.Russian, Tags.Source.Aggregator);
+        super('mangalib', 'MangaLib', 'https://mangalib.org', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Media.Manga, Tags.Language.Russian, Tags.Source.Aggregator);
     }
 
     public override get Icon() {
         return icon;
     }
 
+    public override async Initialize(): Promise<void> {
+        const { data: { imageServers } } = await FetchJSON<APIResult<ImageServers>>(new Request(new URL('constants?fields[]=imageServers', this.apiUrl)));
```
ronny1982, line 74: Consider adding a `./` prefix for URL paths to express a relative location

ronny1982: Other places as well

web/src/engine/websites/MangaLib.ts
```diff
@@ -1,46 +1,125 @@
...
+        this.imageServer = imageServers.find(server => server.id === 'main' && server.site_ids.includes(this.siteId));
```
ronny1982, line 75: How often are these servers changing?
If they almost never change, they may be hardcoded.
If they change daily, users may have to restart the application to update these servers since this is done in the `Initialize` method.

web/src/engine/websites/MangaLib.ts
```diff
@@ -1,46 +1,125 @@
...
+        const mangaSlug = new URL(url).pathname.split('/').at(-1);
+        const { data: { slug_url, rus_name } } = await FetchJSON<APIResult<APIManga>>(new Request(new URL(`manga/${mangaSlug}`, this.apiUrl)));
+        return new Manga(this, provider, slug_url, rus_name);
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const mangaList: Manga[] = [];
+        for (let page = 1, run = true; run; page++) {
+            const url = new URL(`manga?page=${page}&site_id[]=${this.siteId}`, this.apiUrl);
+            const { data, meta: { has_next_page } } = await this.mangasTaskPool.Add(() => FetchJSON<APIResult<APIManga[]>>(new Request(url)), Priority.Low);
```
ronny1982, line 92: When setting the concurrency for `mangasTaskPool` to anything different than `1`, it will not work as intended.
Maybe choose more appropriate name to describe the behavior e.g., `mangasSequentialTaskPool`, `mangasNonCurrentTaskPool`, ...

web/src/engine/websites/MangaLib.ts
```diff
@@ -1,46 +1,125 @@
...
+            mangaList.push(...data.map(manga => new Manga(this, provider, manga.slug_url, manga.rus_name.trim())));
+            run = has_next_page;
```
ronny1982, line 94: No need for the `run` variable, may directly leave the loop `if(!has_next_page) break;`
The `run` variable is used for loops with ternary operator to avoid `if ... else ...` blocks.

web/src/engine/websites/MangaLib.ts
```diff
@@ -1,46 +1,125 @@
...
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const { data } = await FetchJSON<APIResult<APIChapter[]>>(new Request(new URL(`manga/${manga.Identifier}/chapters`, this.apiUrl)));
+        return data.reduce((accumulator: Chapter[], chapter) => {
+            let baseTitle = `Том ${chapter.volume} Глава ${chapter.number}`;
+            baseTitle += chapter.name ? ` - ${chapter.name}` : '';
+            const chapters = chapter.branches.map(branch => {
+                const teamName = chapter.branches_count > 1 ? `[${branch.teams.at(0).name}]` : '';
+                const chapterId = JSON.stringify({
```
ronny1982, line 106: Suggestion to use URL path name including the corresponding search parameters, so it can be directly consumed in `FetchPages` instead of using intermediate JSON serialization.

web/src/engine/websites/MangaLib.ts
```diff
@@ -1,46 +1,125 @@
...
+                return new Chapter(this, manga, chapterId, [baseTitle, teamName].join(' ').trim());
+            });
+            accumulator.push(...chapters);
+            return accumulator;
+        }, []);
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const { branch_id, number, volume } = JSON.parse(chapter.Identifier) as ChapterID;
+        const url = new URL(`manga/${chapter.Parent.Identifier}/chapter?number=${number}&volume=${volume}`, this.apiUrl);
```
ronny1982, line 120: The path name and search parameters can already be composed in `FetchChapters` and provided as chapter identifier.

web/src/engine/websites/MangaLib.ts
```diff
@@ -1,46 +1,125 @@
...
+    branch_id: string,
+    number: string,
+    volume: string
+}
+
+@Common.ImageAjax(true)
 export default class extends DecoratableMangaScraper {
+    private readonly apiUrl = 'https://api.mangalib.me/api/';
+    private readonly siteId = 1;
+    private imageServer: ImageServer = undefined;
```
ronny1982, line 62: May just use type string (url) since `ImageServer.id` and `ImageServer.site_ids` are not used in `FetchImages`.

web/src/engine/websites/MangaLib_e2e.ts
```diff
@@ -1,24 +1,47 @@
...
+        size: 314_272,
+        type: 'image/webp'
     }
 };
 
-new TestFixture(config).AssertWebsite();
\ No newline at end of file
+new TestFixture(configWithBranches).AssertWebsite();
+
+const configWithoutBranches = {
```
ronny1982, line 26: May directly embed this config in `TestFixture` constructor to avoid intermediate variable.
When removing the variable I would suggest to add a comment describing the test case.

web/src/engine/websites/MangaLib_e2e.ts
```diff
@@ -1,24 +1,47 @@
 ﻿import { TestFixture } from '../../../test/WebsitesFixture';
 
-const config = {
+const configWithBranches = {
```
ronny1982, line 3: May directly embed this config in `TestFixture` constructor to avoid intermediate variable.
When removing the variable I would suggest to add a comment describing the test case.

ronny1982 (approved):

web/src/engine/websites/MangaLib_e2e.ts
MikeZeDev: Yeah, it has not being updated since we started to changes tests.

web/src/engine/websites/MangaLib.ts
MikeZeDev: When do you think i should update them? Make a request at the beginning of `FetchPages` ?

web/src/engine/websites/MangaLib.ts
MikeZeDev: I can build a pathname if you want, but it wouldnt be the same one than the website. There is language code in pathname and i fear other Libgroup websites may use multiple language. Well, FetchPages dont need language 🤷

web/src/engine/websites/MangaLib.ts
MikeZeDev: Aaah i see what you meant. mb

PR #883, opened 2024-11-19
AsuraScans: fix getting chapters & pages

MikeZeDev:
Closes https://github.com/manga-download/haruneko/issues/884

### DIFF
--- a/web/src/engine/websites/AsuraScans.ts
+++ b/web/src/engine/websites/AsuraScans.ts
@@ -20,16 +20,12 @@ function MangaInfoExtractor(anchor: HTMLAnchorElement) {
     };
 }
 
-type JSONChapters = {
-    chapters: {
-        name: number
-    }[]
+type JSONChapter = {
+    name: number
 }
 
-type JSONPages = {
-    pages: {
-        url: string
-    }[]
+type JSONPage = {
+    url: string;
 }
 
 @Common.MangasMultiPageCSS('/series?page={page}', 'div.grid a', 1, 1, 0, MangaInfoExtractor)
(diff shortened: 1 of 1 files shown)


web/src/engine/websites/AsuraScans.ts
```diff
@@ -47,8 +58,40 @@ export default class extends DecoratableMangaScraper {
     }
 
     public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
```
ronny1982, line 60: The chapters are directly available as HTML, any reason why extracting these from the scripts which is likely much more error prone?

image

web/src/engine/websites/AsuraScans.ts
```diff
@@ -47,8 +58,40 @@ export default class extends DecoratableMangaScraper {
...
-        const chapters = await FetchCSS<HTMLAnchorElement>(new Request(new URL(manga.Identifier, this.URI)), 'div.scrollbar-thumb-themecolor a.block');
-        return chapters.map(chapter => new Chapter(this, manga, [manga.Identifier, 'chapter', chapter.pathname.match(/(\d+)+$/)[1]].join('/'), chapter.textContent.trim()));
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL(`${manga.Identifier}`, this.URI)), 'script:not([src])');
+        const { chapters } = this.FindJSONObject<JSONChapters>(scripts, /is_early_access/, 'chapters');
+        return chapters.map(chapter => new Chapter(this, manga, [manga.Identifier, 'chapter', chapter.name.toString()].join('/'), `Chapter ${chapter.name}`));
     }
 
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
```
ronny1982, line 66: How about extracting image links with a delayed browser script instead?
Would be possible to avoid the fragile parsing of Next.js client-side rendering code.

web/src/engine/websites/AsuraScans.ts
```diff
@@ -47,8 +58,40 @@ export default class extends DecoratableMangaScraper {
...
+        if (scripts && scriptRegex) {
+            const script = scripts.find(script => scriptRegex.test(script.text))?.text;
+            if (!script) return undefined;
+            //script are like self.__next_f.push([1,"
+            const json = JSON.parse(script.substring(script.indexOf(',"') + 1, script.length - 2));// to remove trailing )]
+            currentElement = JSON.parse(json.substring(json.indexOf(':') + 1));
+            return this.FindJSONObject<T>(undefined, undefined, keyName, currentElement);
+        }
+
+        if (!currentElement) return undefined;
```
ronny1982, line 85: Can this condition even be met, seems redundant to me?

web/src/engine/websites/AsuraScans.ts
```diff
@@ -47,8 +58,40 @@ export default class extends DecoratableMangaScraper {
...
+        if (currentElement[keyName]) {
+            return currentElement;
+        }
+        let result = undefined;
+        for (let i in currentElement) {
+            if (result) break;
+            if (typeof currentElement[i] === 'object')
+                result = result ?? this.FindJSONObject<T>(undefined, undefined, keyName, currentElement[i]);
```
ronny1982, line 93: Return directly as soon as a result is found

web/src/engine/websites/AsuraScans.ts
```diff
@@ -47,8 +58,40 @@ export default class extends DecoratableMangaScraper {
...
+            const json = JSON.parse(script.substring(script.indexOf(',"') + 1, script.length - 2));// to remove trailing )]
+            currentElement = JSON.parse(json.substring(json.indexOf(':') + 1));
+            return this.FindJSONObject<T>(undefined, undefined, keyName, currentElement);
+        }
+
+        if (!currentElement) return undefined;
+        if (currentElement[keyName]) {
+            return currentElement;
+        }
+        let result = undefined;
```
ronny1982, line 89: No need for outer variable declaration when returning directly from the loop.

web/src/engine/websites/AsuraScans.ts
```diff
@@ -47,8 +58,40 @@ export default class extends DecoratableMangaScraper {
...
 
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL(`${chapter.Identifier}`, this.URI)), 'script:not([src])');
+        const { pages } = this.FindJSONObject<JSONPages>(scripts, /chapterName/, 'pages');
+        return pages
+            .filter(page => excludes.none(pattern => pattern.test(page.url)))
+            .map(page => new Page(this, chapter, new URL(page.url)));
+    }
+
+    private FindJSONObject<T>(scripts: HTMLScriptElement[], scriptRegex: RegExp, keyName: string, currentElement = undefined): T {
```
ronny1982, line 74: Split responsibilities => first method for parsing scripts to objects, second method for searching key in objects

ronny1982 (approved):

web/src/engine/websites/AsuraScans.ts
MikeZeDev: Its precisely because those selectors are changing very often that i would rather extract chapters from json structure and that scrapers have troubles. The selectors on React websites are very hard to chose the class names are too generic.

web/src/engine/websites/AsuraScans.ts
MikeZeDev: You told me to use one single method in another plugin, now you want two 😢

ronny1982: Don't know about the other comment, but in this specific case the parameters `scripts` and `scriptRegex` are only used once for parsing and become redundant (are used as flags for conditionally excluding part of the business logic in consecutive calls) which indicates a violation of the single responsibility principle.

It would be a different story if there was no recursive call.

ronny1982: Found it in #821

The remark was related to combining the repetition of multiple calls.

```javascript
// somewhere
const tmp1 = A();
const res1 = B(tmp1);

// somewhere else
const tmp2 = A();
const res2 = B(tmp2);
```

Which was addressed by merging the business logic of `A()` and `B()` into a single replacement function, but could have also been addressed by grouping these calls into another method.
```javascript

C() {
  const tmp = A();
  return B(tmp);
}

// somewhere
const res1 = C();

// somewhere else
const res2 = C();
```

To be fair the final implementation of the single replacement function (which is also presented in this PR) was merged into `master` directly without any further review/feedback.

ronny1982: I made some adjustments in CoroCoro which could be used for reference
https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/CoroCoro.ts#L86

MikeZeDev: Sorry for this misundernstanding.

PR #896, opened 2024-11-22
Allow Uint8Array as GetTypedData parameters

MikeZeDev:
* npm run checks now gives errors because `Common.GetTypedData` wants STRICT `ArrayBuffer` as parameter.

* This appends when 

1. We get picture as `ArrayBuffer`,  create an `Uint8Array` from it, decrypt the picture using that `Uint8Array`.
2. We `return Common.GetTypedData(decrypted_Uint8Array)`

I.e : NicoNicoSeiga.

Its stupid to convert `decrypted_Uint8Array` to an ArrayBuffer AGAIN because `GetTypedData` will convert it to `Uint8Array` anyway.


MikeZeDev:
Superseded by https://github.com/manga-download/haruneko/commit/07be4b49b50d7fb338150695f9676ac04a3d8eac

PR #903, opened 2024-11-25
Add Grouple Websites

MikeZeDev:


web/src/engine/websites/AllHentai.ts
```diff
@@ -0,0 +1,27 @@
+import { Tags } from '../Tags';
+import icon from './AllHentai.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+import * as Grouple from './decorators/Grouple';
+//import { Text } from '../SettingsManager';
+//import { WebsiteResourceKey as R } from '../../i18n/ILocale';
+
+// TODO: add url setting
```
ronny1982, line 9: Can't the URL be updated during`Initialization`?

web/src/engine/websites/MintManga_e2e.ts
```diff
@@ -0,0 +1,26 @@
...
+        id: 'mintmanga',
+        title: 'MintManga'
+    },
+    container: {
+        url: 'https://2.mintmanga.one/blagoslovenie_nebojitelei__A3b0f6',
+        id: '/blagoslovenie_nebojitelei__A3b0f6',
+        title: 'Благословение небожителей',
+    }, /* Need to be logged, and chapter link change with user id anyway
+    child: {
+        id: '/blagoslovenie_nebojitelei__A3b0f67/vol8/98',
```
ronny1982, line 14: Parameter `?mtr=1` is missing

web/src/engine/websites/AllHentai_e2e.ts
```diff
@@ -0,0 +1,26 @@
...
+    },
+    container: {
+        url: 'https://20.allhen.online/nochiu_ona_nimfomanka',
+        id: '/nochiu_ona_nimfomanka',
+        title: 'Ночью она нимфоманка'
+    },
+/*   Need to be logged, and chapter link change with user id anyway
+
+    child: {
+        id: '/nochiu_ona_nimfomanka/vol1/4',
```
ronny1982, line 16: Parameter `?mtr=1` is missing

ronny1982 (approved):
Consider adding `Grouple_e2e.ts` including all websites which are using this template, for faster verification when changes are made to the template.

web/src/engine/websites/AllHentai_e2e.ts
MikeZeDev: It doesnt matter, as the test will NEVER be good. On this website chapter will ALWAYS has user token as parameter and there is none when not logged. Fine, i'll add it

web/src/engine/websites/MintManga_e2e.ts
MikeZeDev: Same reasoning than above.

PR #943, opened 2024-12-20
DelitoonBase: move image unscrambling in base class

MikeZeDev:
Its not used in Bomtoon only now.

Reference : https://github.com/manga-download/hakuneko/issues/7617

Also, use a class parameter for pages endpoint.


MikeZeDev:
> I’m not sure if I can reply here, but [BomtoonCN] has the same issue. I really hope it can be fixed together. Thank you

Rest assured, the point of this PR is to make the unscrambling part initially done for Bomtoon available for all similar websites.
DelitoonBase being "the base template".

PR #962, opened 2025-01-03
Ganma : use new api (GraphQL)

MikeZeDev:


ronny1982 (approved):
Maybe the query fields can be reduced, they seem to request various fields that may not be relevant for HakuNeko.

MikeZeDev:
> Maybe the query fields can be reduced, they seem to request various fields that may not be relevant for HakuNeko.

There is only isPurchased that is not used and i kept it just in case.

MikeZeDev:
'OnlyPersistedQueryIsAllowed" error. Cant use Graphql anymore apparently. F**** them.

Issue #964, opened 2025-01-04
Vortex Scans  complete chapter images not loading


Zoroam:
For the last few days, the Vortexscans.org connector is just opening 3 images of every chapter. It downloads just 3 images of the chapter and doesn't download the rest of the images. Issue is happening with all the comics available on this connector.

MikeZeDev:
Meh, images are lazy loaded only when you manually scroll the page...
And its NextJS chunks bullshit.

Not sure its gonna be ez to fix this ******

PR #965, opened 2025-01-04
Vortex Scans : fix getting images

MikeZeDev:
and update Hivescans tests

Closes https://github.com/manga-download/haruneko/issues/964

### DIFF
--- a/web/src/engine/websites/VortexScans.ts
+++ b/web/src/engine/websites/VortexScans.ts
@@ -1,6 +1,12 @@
 import { Tags } from '../Tags';
+import { FetchCSS } from '../platform/FetchProvider';
+import { Page, type Chapter } from '../providers/MangaPlugin';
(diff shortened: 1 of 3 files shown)


web/src/engine/websites/VortexScans.ts
```diff
@@ -11,4 +17,12 @@ export default class extends VTheme {
     public override get Icon() {
         return icon;
     }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const scripts = await FetchCSS<HTMLScriptElement>(new Request(new URL(chapter.Identifier, this.URI)), 'script:not([src])');
```
ronny1982, line 22: I think it is not necessary to filter the scripts, the RegExp can probably directly be used on the plain HTML text `await (await Fetch()).text()`

ronny1982 (approved):

PR #970, opened 2025-01-10
ManjaNoon : no more MangaStream

MikeZeDev:


web/src/engine/websites/Manjanoon.ts
```diff
@@ -1,20 +1,26 @@
...
 
     public constructor() {
-        super('manjanoon', 'Manjanoon', 'https://vrnoin.site', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Media.Manga, Tags.Language.Arabic, Tags.Source.Scanlator);
+        super('manjanoon', 'Manjanoon', 'https://manjanoon.art', Tags.Media.Manhwa, Tags.Media.Manhua, Tags.Media.Manga, Tags.Language.Arabic, Tags.Source.Scanlator);
     }
     public override get Icon() {
         return icon;
     }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
```
ronny1982, line 20: May use `PagesSinglePageJS` instead

ronny1982 (approved):

web/src/engine/websites/Manjanoon.ts
MikeZeDev: Irk, but i'd rather refrain opening a browser window and executing foreign code when i can.
Plus, doing this way is lighter i guess.

PR #975, opened 2025-01-16
Add FlatManga template

MikeZeDev:
template only

### DIFF
--- a/web/src/engine/websites/templates/FlatManga.ts
+++ b/web/src/engine/websites/templates/FlatManga.ts
@@ -2,6 +2,7 @@
 import { DecoratableMangaScraper, type Chapter, type MangaScraper } from "../../providers/MangaPlugin";
 import { Page } from "../../providers/MangaPlugin";
 import * as Common from '../decorators/Common';
+
 export function MangaLabelExtractor(element: HTMLElement) {
     return CleanTitle(element.getAttribute('text') ? element.getAttribute('text') : element.textContent);
 }
--- a/web/src/engine/websites/templates/FlatManga.ts
+++ b/web/src/engine/websites/templates/FlatManga.ts
@@ -1,0 +1,110 @@
+﻿import { AddAntiScrapingDetection, FetchRedirection } from "../../platform/AntiScrapingDetection";
+import { FetchCSS } from "../../platform/FetchProvider";
+import { DecoratableMangaScraper, type Chapter, type MangaScraper } from "../../providers/MangaPlugin";
+import { Page } from "../../providers/MangaPlugin";
+import * as Common from '../decorators/Common';
+
+AddAntiScrapingDetection(async (invoke) => {
+    const result = await invoke<boolean>(`document.documentElement.innerHTML.includes('ct_anti_ddos_key')`);// Sample => Mangagun, NicoManga, Rawinu, Weloma, WeloveManga
+    if (result) {
+        await new Promise(resolve => setTimeout(resolve, 3000));
+        return FetchRedirection.Automatic;
+    } else return undefined;
+
+});
+
+export function MangaLabelExtractor(element: HTMLElement) {
+    return CleanTitle(element.getAttribute('text') ? element.getAttribute('text') : element.textContent);
+}
+
+export function MangaExtractor(anchor: HTMLAnchorElement) {
+    return {
+        id: anchor.pathname,
+        title: CleanTitle(anchor.getAttribute('text') ? anchor.text : anchor.textContent)
+    };
+}
+
+export function CleanTitle(title: string): string {
+    title = title.replace(/\s*[~\-―〜]\s*RAW\s*\(MANGA\)\s*$/i, '').trim();
+    title = title.replace(/\s*[~\-―〜]\s*(MANGA)?\s*RAW\s*$/i, '').trim();
+    title = title.replace(/\(raw\)/i, '').trim();
+    return title.replace(/\(manga\)/i, '').trim();
+}
+
+export function PageLinkExtractor<E extends HTMLImageElement>(this: MangaScraper, element: E): string {
+    let page = element.dataset.aload || element.dataset.src || element.dataset.srcset || element.dataset.original || element.dataset.pagespeedLazySrc || element.src;
+    try {
+        page = window.atob(page);
+    } catch { }
+    return page.replace(/\n/g, '');
+
+}
+
+export const pathSinglePageManga = '/manga-list.html?listType=allABC';
+const pathMultiPageManga = '/manga-list.html?page={page}';
+
+export const queryMangaTitle = [
+    'li:last-of-type span[itemprop="name"]',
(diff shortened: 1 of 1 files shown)


web/src/engine/websites/templates/FlatManga.ts
```diff
@@ -0,0 +1,110 @@
+﻿import { AddAntiScrapingDetection, FetchRedirection } from "../../platform/AntiScrapingDetection";
+import { FetchCSS } from "../../platform/FetchProvider";
+import { DecoratableMangaScraper, type Chapter, type MangaScraper } from "../../providers/MangaPlugin";
+import { Page } from "../../providers/MangaPlugin";
+import * as Common from '../decorators/Common';
+
+AddAntiScrapingDetection(async (invoke) => {
+    const result = await invoke<boolean>(`document.documentElement.innerHTML.includes('ct_anti_ddos_key')`);// Sample => Mangagun, NicoManga, Rawinu, Weloma, WeloveManga
```
manga-download-owner, line 8: Prefer CSS selector instead of matching text since this would be less error prone

ronny1982, line 8: Prefer CSS selector instead of matching text since this would be less error prone

web/src/engine/websites/templates/FlatManga.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (result) {
+        await new Promise(resolve => setTimeout(resolve, 3000));
```
ronny1982, line 10: I don't understand the delay at this point.
I think this is not necessary and may lead to errors in case the website navigates within these 3 seconds an the automatic redirection may not be performed afterwards.

web/src/engine/websites/templates/FlatManga.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export function CleanTitle(title: string): string {
+    title = title.replace(/\s*[~\-―〜]\s*RAW\s*\(MANGA\)\s*$/i, '').trim();
+    title = title.replace(/\s*[~\-―〜]\s*(MANGA)?\s*RAW\s*$/i, '').trim();
+    title = title.replace(/\(raw\)/i, '').trim();
+    return title.replace(/\(manga\)/i, '').trim();
+}
+
+export function PageLinkExtractor<E extends HTMLImageElement>(this: MangaScraper, element: E): string {
+    let page = element.dataset.aload || element.dataset.src || element.dataset.srcset || element.dataset.original || element.dataset.pagespeedLazySrc || element.src;
```
ronny1982, line 35: The `dataset.aload` and `dataset.pagespeedLazySrc` seems to be very website specific, should these really be part of the template?

web/src/engine/websites/templates/FlatManga.ts
```diff
@@ -0,0 +1,110 @@
...
+        page = window.atob(page);
+    } catch { }
+    return page.replace(/\n/g, '');
+
+}
+
+export const pathSinglePageManga = '/manga-list.html?listType=allABC';
+const pathMultiPageManga = '/manga-list.html?page={page}';
+
+export const queryMangaTitle = [
```
ronny1982, line 46: Are all of these CSS selectors really part of the template, or are there some website specific?
If there are like 10 websites and each has a different CSS selector, than these should not be added to the template.
Only the shared CSS selectors should be added.

web/src/engine/websites/templates/FlatManga.ts
```diff
@@ -0,0 +1,110 @@
...
+    'li:last-of-type span[itemprop="name"]',
+    'ul.manga-info h3',
+    'ul.manga-info h1',
+].join(',');
+
+export const queryMangas = [
```
ronny1982, line 52: Are all of these CSS selectors really part of the template, or are there some website specific?
If there are like 10 websites and each has a different CSS selector, than these should not be added to the template.
Only the shared CSS selectors should be added.

web/src/engine/websites/templates/FlatManga.ts
```diff
@@ -0,0 +1,110 @@
...
+    'span[data-toggle="mangapop"] a',
+    'div.media h3.media-heading a',
+    'div.container a[data-toggle="mangapop"]',
+    'div.card-body div.series-title a'
+].join(',');
+
+const queryChapters = [
```
ronny1982, line 59: Are all of these CSS selectors really part of the template, or are there some website specific?
If there are like 10 websites and each has a different CSS selector, than these should not be added to the template.
Only the shared CSS selectors should be added.

web/src/engine/websites/templates/FlatManga.ts
```diff
@@ -0,0 +1,110 @@
...
+    'div#tab-chapper div#list-chapters span.title a.chapter'
+].join(',');
+
+export const queryPages = [
+    'img.chapter-img',
+    'div#chapter-content img',
+    'div.chapter-content img'
+].join(',');
+
+export const chapterScript = `
```
ronny1982, line 71: Unused export, should be added when needed

web/src/engine/websites/templates/FlatManga.ts
```diff
@@ -0,0 +1,110 @@
...
+    [...document.querySelectorAll('ul.list-chapters > a')].map(chapter => {
+        return {
+            id: chapter.pathname,
+            title : chapter.title.trim()
+        };
+    });
+`;
+
+export const pageScript = `
```
ronny1982, line 80: Unused export, should be added when needed

ronny1982 (changes requested):

web/src/engine/websites/templates/FlatManga.ts
MikeZeDev: Anti Ddoss is delaying the redirection after check is done.

web/src/engine/websites/templates/FlatManga.ts
MikeZeDev: I understand the concern but its too much a hassle to remove for i have to reflect changes on another branch where all the websites are and where its needed.

I made sure no unnecessary stuff is exported already. You cant see that because you wanted smaller PR 🤷

web/src/engine/websites/templates/FlatManga.ts
MikeZeDev: I understand the concern but its too much a hassle to remove for i have to reflect changes on another branch where all the websites are and where its needed.

I made sure no unnecessary stuff is exported already. You cant see that because you wanted smaller PR 🤷

web/src/engine/websites/templates/FlatManga.ts
MikeZeDev: Most websites are using either 'ul.manga-info h3' either 'ul.manga-info h1'. I can prune the other one but removing more would mean unnecessary rewrite in a couple of websites.

ronny1982 (approved):

PR #976, opened 2025-01-17
Add KLMangaSH

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7669


web/src/engine/websites/KLMangaSh.ts
```diff
@@ -0,0 +1,90 @@
...
+        for (let run = true; run;) {
+            const request = new Request(this.zingParams.ajax_url, {
+                method: 'POST',
+                body: params,
+                headers: {
+                    'Content-type': 'application/x-www-form-urlencoded; charset=UTF8',
+                    'X-Requested-With': 'XMLHttpRequest',
+                }
+            });
+            let { img_index, mes, going } = await FetchJSON<PageResult>(request);
```
ronny1982, line 83: The `FetchPages` method must avoid making too many requests.
Either parallelize the requests, or better move the logic to `FetchImage`.

web/src/engine/websites/KLMangaSh.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+function CleanTitle(title: string): string {
+    return title.replace(/\(Raw.*Free\)/i, '').trim();
+}
+
+function MangaLabelExtractor(element: HTMLElement) : string{
+    return CleanTitle(element.textContent);
+}
+function MangaExtractor(anchor: HTMLAnchorElement) {
```
ronny1982, line 25: Missing blank line

ronny1982 (approved):

web/src/engine/websites/KLMangaSh.ts
```diff
@@ -0,0 +1,90 @@
...
+
+@Common.MangaCSS(/^{origin}\/manga-raw\/[^/]+\/$/, 'div.container div.z-single-mg h1.name', MangaLabelExtractor)
+@Common.MangasMultiPageCSS('/page/{page}/', 'div.grid-of-mangas h2.name a', 1, 1, 0, MangaExtractor)
+@Common.ChaptersSinglePageCSS('div.chapter-box a', ChapterExtractor)
+@Common.ImageAjax(true)
+export default class extends DecoratableMangaScraper {
+    private zingParams: ZingParams;
+
+    public constructor() {
+        super('klmangash', 'KLManga(.sh)', 'https://klmanga.sh', Tags.Language.Japanese, Tags.Source.Aggregator);
```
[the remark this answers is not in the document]
MikeZeDev: There IS one. Its just not added yet. Its part of Flatmanga.

web/src/engine/websites/KLMangaSh.ts
MikeZeDev: Ikr, but we DONT KNOW the real number of pages so both of your solutions are impossible.
The real number of pages is known once the website api return going = 0 ie. after the loop.

Crappy website coding.

ronny1982:
The filenames should have the same casing, either `KLMangaSh` or `KLMangash`

web/src/engine/websites/KLMangaSh.ts
MikeZeDev: I've checked rawlazy before submitting that PR, since i saw the similarities. Rawlazy is simply using lazy loading, with one img element per page. Here we have nothing afaik.

MikeZeDev: I mean there is a dirty way. Make the request once, you get first two pages.
Then we have the page folder.

`hoster.com/path/number.extension.`

Since the pages urls follow a pattern we could send HEAD requests while incrementing number.

ronny1982: May find the last page by using divide and conquer algorithm, within 6 requests it would be possible to find the last API call which provides a non empty content (assuming there are < 128 pages).

You may leave as is for now, may take a look into it later

PR #985, opened 2025-01-28
Add MyComic

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7672


web/src/engine/websites/MyComic.ts
```diff
@@ -0,0 +1,43 @@
...
+        super('mycomic', 'MyComic', 'https://mycomic.com', Tags.Media.Manga, Tags.Media.Manhua, Tags.Media.Manhwa, Tags.Media.Comic, Tags.Language.Chinese, Tags.Source.Aggregator);
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const result: Chapter[] = [];
+        const matches = await FetchRegex(new Request(new URL(manga.Identifier, this.URI)), /chapters\s*:\s*(\[.*\])/g);
```
ronny1982, line 35: The name `match` is confusing as it could be inferred by a developer to contain a list of regex result objects, but in fact this is a list of strings. Consider using a more appropriate variable name.

ronny1982 (approved):

PR #989, opened 2025-01-31
disable Chrome background throttling

MikeZeDev:
Users has been complaining that "downloads are slower in background".

This is using electron window parameter & nwjs command line switch.

The impact needs to be determined.

### DIFF
--- a/app/nw/scripts/build-app.mjs
+++ b/app/nw/scripts/build-app.mjs
@@ -26,7 +26,7 @@ const manifest = {
         'https://app.hakuneko.download/*',
         `${new URL(pkgConfig.url).origin}/*`,
     ],
-    'chromium-args': '--disable-raf-throttling',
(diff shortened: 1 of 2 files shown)


ronny1982:
I think a more appropriate approach would be to move the background timings to a web-worker.

MikeZeDev:
Wont this require a new release?

ronny1982:
> Wont this require a new release?

This PR requires a new release, the web-worker approach does not

MikeZeDev:
> > Wont this require a new release?
> 
> This PR requires a new release, the web-worker approach does not

Exactly what i was asking, thanks :)

ronny1982:
Oh the fix for Electron is in the wrong file, should be in [Main.ts](https://github.com/manga-download/haruneko/blob/master/app/electron/src/Main.ts#L61-L69)

PR #990, opened 2025-01-31
Baozimh : fix getting pages

MikeZeDev:
* Fix endless loop on last chapter
* Use alternate CDN to remove watermark

### DIFF
--- a/web/src/engine/websites/Baozimh.ts
+++ b/web/src/engine/websites/Baozimh.ts
@@ -1,6 +1,6 @@
 import { Tags } from '../Tags';
 import icon from './Baozimh.webp';
-import { DecoratableMangaScraper, type MangaPlugin, Manga, Page, type Chapter } from '../providers/MangaPlugin';
+import { type Chapter, DecoratableMangaScraper, Page, type MangaPlugin, Manga } from '../providers/MangaPlugin';
 import * as Common from './decorators/Common';
 import { FetchHTML, FetchJSON } from '../platform/FetchProvider';
 
--- a/web/src/engine/websites/Baozimh.ts
+++ b/web/src/engine/websites/Baozimh.ts
@@ -52,8 +52,9 @@ export default class extends DecoratableMangaScraper {
         const chapterSlot = chapterPageUri.searchParams.get('chapter_slot');
         for (let hasNextPage = true, pageIndex = 2; hasNextPage; pageIndex++) {
             const doc = await FetchHTML(new Request(chapterPageUri));
-            pagesList.push(...[...doc.querySelectorAll<HTMLElement>('.comic-contain amp-img')]
-                .map(element => new Page(this, chapter, new URL(this.ReplaceCDN(element.dataset.src)))));
+            const pages = [...doc.querySelectorAll<HTMLElement>('.comic-contain amp-img')]
+                .map(element => new Page(this, chapter, new URL(this.ReplaceCDN(element.dataset.src))));
+            pagesList.push(...pages);
             chapterPageUri = new URL([...doc.querySelectorAll<HTMLAnchorElement>('div.comic-chapter div.next_chapter a')].at(-1)?.pathname, this.URI);
             hasNextPage = new RegExpSafe(`/${sectionSlot}_${chapterSlot}_${pageIndex}\\.html?$`, 'i').test(chapterPageUri.href);
         }
--- a/web/src/engine/websites/Baozimh.ts
+++ b/web/src/engine/websites/Baozimh.ts
@@ -47,15 +47,15 @@ export default class extends DecoratableMangaScraper {
 
     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
         const pagesList: Page[] = [];
-        let uri = new URL(chapter.Identifier, this.URI);
-        const sectionSlot = uri.searchParams.get('section_slot');
-        const chapterSlot = uri.searchParams.get('chapter_slot');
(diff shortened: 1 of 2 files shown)


web/src/engine/websites/Baozimh_e2e.ts
```diff
@@ -16,9 +16,7 @@ const config = {
     },
     entry: {
         index: 5,
-        size: 164_832,
+        size: 151_930,
```
ronny1982, line 19: The image is now smaller, therefore it seems replacing the CDN leads to lower quality images.

web/src/engine/websites/Baozimh.ts
```diff
@@ -46,18 +46,21 @@ export default class extends DecoratableMangaScraper {
...
-            uri = new URL([...data.querySelectorAll<HTMLAnchorElement>('div.comic-chapter div.next_chapter a')].at(-1)?.pathname, this.URI);
-            run = linkRegex.test(uri.href);
+        const pagesList: Page[] = [];
+        let chapterPageUri = new URL(chapter.Identifier, this.URI);
+        const sectionSlot = chapterPageUri.searchParams.get('section_slot');
+        const chapterSlot = chapterPageUri.searchParams.get('chapter_slot');
+        for (let hasNextPage = true, pageIndex = 2; hasNextPage; pageIndex++) {
+            const doc = await FetchHTML(new Request(chapterPageUri));
+            pagesList.push(...[...doc.querySelectorAll<HTMLElement>('.comic-contain amp-img')]
+                .map(element => new Page(this, chapter, new URL(this.ReplaceCDN(element.dataset.src)))));
```
ronny1982, line 56: Five closing brackets at the end could be an indicator that there is too much going on in this single instruction (which makes it a little hard to read).

web/src/engine/websites/Baozimh.ts
```diff
@@ -46,18 +46,21 @@ export default class extends DecoratableMangaScraper {
...
-            const data = await FetchHTML(new Request(uri));
-            const pages = [...data.querySelectorAll('.comic-contain amp-img.comic-contain__item')];
-            pagesList.push(...pages.map(element => new Page(this, chapter, new URL(element.getAttribute('src'), this.URI))));
-            uri = new URL([...data.querySelectorAll<HTMLAnchorElement>('div.comic-chapter div.next_chapter a')].at(-1)?.pathname, this.URI);
-            run = linkRegex.test(uri.href);
+        const pagesList: Page[] = [];
+        let chapterPageUri = new URL(chapter.Identifier, this.URI);
+        const sectionSlot = chapterPageUri.searchParams.get('section_slot');
+        const chapterSlot = chapterPageUri.searchParams.get('chapter_slot');
+        for (let hasNextPage = true, pageIndex = 2; hasNextPage; pageIndex++) {
```
ronny1982, line 53: I don't see a need for this loop, all images could be extracted from a [single HTML](https://www.czmanga.com/comic/chapter/wotianmingdafanpai-tianmingfanpaiyuanzhuhuizhudongman/0_53.html) file (which can be acquired by following the redirect of the chapter link).

ronny1982 (changes requested):

web/src/engine/websites/Baozimh_e2e.ts
MikeZeDev: I dont think the resolution is lower. The thing is : normal cdn has big watermarks 🤷

web/src/engine/websites/Baozimh.ts
MikeZeDev: I would like to use Common.FetchPageSinglePageCSS in a loop until i realised it wouldnt work as chapters pages are on another host.

web/src/engine/websites/Baozimh.ts
MikeZeDev: Did you see the "1/3"? A page only display 50 images max. In the past we had complains that is was multipaged and it doesnt seems to have changed.

If you check the [first page](https://www.czmanga.com/comic/chapter/wotianmingdafanpai-tianmingfanpaiyuanzhuhuizhudongman/0_53.html) and the [second page](https://www.czmanga.com/comic/chapter/wotianmingdafanpai-tianmingfanpaiyuanzhuhuizhudongman/0_53_2.html)  you will see that the pictures are not the same and that second page follow the first.

web/src/engine/websites/Baozimh_e2e.ts
MikeZeDev: Pictures resolutions are the same on both sides. Dont forget they are different (watermark/no watermark), so i assume the size differences are because of compression 🤷

web/src/engine/websites/Baozimh.ts
ronny1982: No i didn't see the other pages due to their bad website design. I guess as an average user I would have missed almost two 3rd of the chapters content.

ronny1982 (approved):

web/src/engine/websites/Baozimh_e2e.ts
ronny1982: I was referring to smaller in size not in resolutions as there was a decrease by 13KB.
The missing watermark seems to explain why the new image is smaller in size.

web/src/engine/websites/Baozimh.ts
MikeZeDev: We first missed it when website was added ;)

PR #1015, opened 2025-02-08
Introduce Background Timers

ronny1982:
Provide alternative timer functions to avoid Chrome background throttling for `setTimeout` and `setInterval`.

This solution utilizes web-workers and therefore comes with some drawbacks such as inter-process communication overhead (for such a minor change) and increased difficulty in automated testing due to web-workers only being available in browsers but not in the NodeJS test environment.

Need to be decided if web-workers shall be used (which will work in all browser-like environments), or the much simpler but less generic hack in NW.js and Electron should be kept as is.


MikeZeDev:
I dont know. I mean, being generic is awesome. If possible we should keep those workers.

I am concerned about : 

1) First and foremost, the obvious question : does it work properly ? Since its a heavy change we have to make sure this is the good solution, and working. I have full confidence in your programing skills but confidence doesn't replace testing.
Do you had the time to test it? Should i test myself too? 🤔 

2) This code is way harder to understand and maintain.

MikeZeDev:
I am gonna make more testing, as the one were perhaps too intensive and not related.

MikeZeDev (approved):

Sheepux:
/LGTM 
it's actually clean to read

Issue #982, opened 2025-01-23
[Bilibili] Cannot read properties of undefined (reading 'map')


RandomGithubUser47:
Comic link: https://manga.bilibili.com/detail/mc35919
Name: 私房照

Error was able to be reproduced on two different computers.

![Image](https://github.com/user-attachments/assets/be5b4e35-6c28-462e-bee2-ef9d3dcf165c)

MikeZeDev:
The made changes again, lmao

PR #1027, opened 2025-02-15
Bilibili: update decryption

MikeZeDev:
Closes https://github.com/manga-download/haruneko/issues/982

https://github.com/manga-download/haruneko-proprietary/commit/763a54a96ca23b501ffe046bb0fa852fcabcd582


web/src/engine/websites/BilibiliManhua.d.ts
```diff
@@ -1,7 +1,8 @@
 import type { Chapter, Manga, Page } from '../providers/MangaPlugin';
 import { DecoratableMangaScraper, type MangaPlugin } from '../providers/MangaPlugin';
-import type { Priority } from '../taskpool/DeferredTask';
+import type { Priority } from '../taskpool/TaskPool';
 export default class extends DecoratableMangaScraper {
+    private readonly mangasTaskPool;
```
ronny1982, line 5: The manga task pool is not used as intended, therefore it will only process a single page at a time.
To utilize the concurrency all tasks should be added at once and awaited with `Promise.all(tasks)`

web/src/engine/websites/BilibiliManhua.js
ronny1982, line 1: Should merge the latest change to avoid fetching each single image with a browser window

ronny1982 (approved):

web/src/engine/websites/BilibiliManhua.d.ts
MikeZeDev: Since we dont have page number i will just remove that.

MikeZeDev: should i just change for  
`  private readonly mangasTaskPool = new TaskPool(1, new RateLimit(4, 1));`

ronny1982: That would work, but may mislead other contributors in thinking the 1 may be changed.
Renaming the field may be more expressive e.g., `mangasSequentialTaskPool` or something like that (maybe Copilot siggests a better name).

PR #1038, opened 2025-02-24
Mangago : fix randomly getting only 5 pages 

MikeZeDev:
https://github.com/manga-download/haruneko-proprietary/commit/b44026f54b1ac90db9637211c0fa757f85298920


ronny1982:
Since there was no test case added, I assume the [current test case](https://www.mangago.me/read-manga/the_dame_in_shining_armor/) would be affected by the change.
However, the current implementation of MangaGo works just fine with the [current test case](https://www.mangago.me/read-manga/the_dame_in_shining_armor/) (chapter 10 shows 66 images)

MikeZeDev:
Of course the tests are "working", as they doesn't even cover gathering chapters. Because chapters domain is random, therefore their id is random too (we use full url as id). 

During testing, i have found that the legacy code still work depending on the generated chapter url pattern.
This PR works with old pattern (where its instant) and new pattern ( paged requests).

MikeZeDev:
this is what i got on chapter 10 with old code 

https://www.youhim.me/chapter/72005/1708354/

image

66 "images" yeah... we got only first five.

ronny1982:
Of course i didn't run the test automatically since it is commented out.
I was referring to the test case which I performed manually and all 66 images showed up.

So he essential information which was not clear to me is the origin. Chapters on `mangago.me` contain all images, while fake domains do not.

I will check this when i got some spare time.

MikeZeDev:
I see what you want to do. Not sure its 100% reliable but its worth a try :)

MikeZeDev:
@ronny1982 your solution seems to works well.

PR #1040, opened 2025-02-26
Bilibili : change url parameters

MikeZeDev:


MikeZeDev:
Changes are simple, user validated that is working.

PR #1051, opened 2025-03-05
Add Senkuro

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7382


ronny1982 (approved):
Approved without review

PR #1058, opened 2025-03-12
Mangadex : fix for removed scanlation group

MikeZeDev:
edge case error in MD database

Issue notified by user and identified by Mangadex contributors: 
https://forums.mangadex.org/threads/error-when-reading-the-chapters-of-one-manga.2170890


web/src/engine/websites/MangaDex.ts
```diff
@@ -40,8 +40,8 @@ type APIChapter = {
     relationships: {
         id: string
         type: string
-        attributes: {
-            name: string
+        attributes?: {
+            name?: string
```
ronny1982, line 44: Both can be `null` attribute and name, or is only attribute nullable?

ronny1982 (approved):

web/src/engine/websites/MangaDex.ts
MikeZeDev: IIRC "attributes" was missing due to a DB error.

> "Cannot read properties of undefined (reading 'name')"

PR #1070, opened 2025-03-23
Port PlotTwistNoFansub

MikeZeDev:


ronny1982 (approved):

MikeZeDev:
What could we do about Codacity error?

ronny1982:
Maybe use `DOMParser` as used for other HTML text with entities
```typescript
new DOMParser().parseFromString(`Meow &lt; Woof`, 'text/html').body.innerText;
```

PR #1080, opened 2025-03-31
Fix Reaperscans inconsistent downloads.

Arch-Storm:
I've found that the reaperscans plugin returns no images for every 5 or so chapters (at least on my end). The reason seems to be that in some cases the server sends a html template string inside the div#content and actually stores the container content inside a #S:1 div instead. Adding that css selector as a possibility returns the plugin to 100% retrieval again.

### DIFF
--- a/web/src/engine/websites/ReaperScans.ts
+++ b/web/src/engine/websites/ReaperScans.ts
@@ -81,7 +81,7 @@ export default class extends DecoratableMangaScraper {
     }
 
     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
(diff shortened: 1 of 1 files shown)


Arch-Storm:
Nvm still not fixed. Will try and find out if this a issue on my end or if theres an actual problem in haruneko

Arch-Storm:
Ok the pr is actually still valid. I just misunderstood how the bundler works and thought it would reflect the web/build changes in the bundled nw app until i realized the bundler just connects to https://app.hakuneko.download instead of bundling the local files.
After going into the dev environment using the local web url my commit fixes the problem again.

PR #1081, opened 2025-03-31
Fix Reaperscans inconsistent downloads.

Arch-Storm:
I've found that the reaperscans plugin returns no images for every 5 or so chapters (at least on my end). The reason seems to be that in some cases the server sends a html template string inside the div#content and actually stores the container content inside a #S:1 div instead. Adding that css selector as a possibility returns the plugin to 100% retrieval again.

(sorry for reopening the pr so many times, i didn't want to merge from my master branch

### DIFF
--- a/web/src/engine/websites/ReaperScans.ts
+++ b/web/src/engine/websites/ReaperScans.ts
@@ -81,7 +81,7 @@ export default class extends DecoratableMangaScraper {
     }
 
     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
-        const pages = await Common.FetchPagesSinglePageCSS.call(this, chapter, 'div#content div.container > div img:not([alt *= "thumb"])', PageLinkExtractor);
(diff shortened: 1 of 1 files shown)


MikeZeDev:
Perhaps you should provide a sample chapter and therefore add a test case in "ReaperScans_e2e.ts". 
I mean, being able to reproduce it by ourselves is a must for maintainers.

Arch-Storm:
> Perhaps you should provide a sample chapter and therefore add a test case in "ReaperScans_e2e.ts".
I mean, being able to reproduce it by ourselves is a must for maintainers.

Would love to do that but the problem is that this doesn't happen on any specific chapters that i can target in unit tests. It seems to happen entirely at random every now and then. I don't see any way to test for this in e2e.

The PR would certainly be a better a solution, but unless it merges soon, this could be a decent stopgap until then.

ronny1982 (approved):

PR #1082, opened 2025-03-31
Improve ComicK api params

Arch-Storm:
I noticed that comick only allows the page and limit to be <=50. but since comick hosts way more than 2500 (50*50) comics on it's site, the fetcher misses a ton of the popular ones. i noticed this because i was trying to download the no 1 most popular manhwa on the site which wasn't found in haruneko. Sorting the api by user_follow_count, shows only the 2500 most popular comics on the site. still not perfect but certainly better then the ones uploaded latest. oh and the limit could be 50 instead of 49 adding another 50 total results.

### DIFF
--- a/web/src/engine/websites/ComicK.ts
+++ b/web/src/engine/websites/ComicK.ts
@@ -91,7 +91,7 @@ export default class extends DecoratableMangaScraper {
 
     private async GetMangasFromPage(page: number, provider: MangaPlugin): Promise<Manga[]> {
         try {
-            const data = await FetchJSON<APIManga[]>(new Request(new URL(`v1.0/search?page=${page}&limit=49`, this.apiUrl)));
+            const data = await FetchJSON<APIManga[]>(new Request(new URL(`v1.0/search?page=${page}&limit=50&sort=user_follow_count`, this.apiUrl)));
             return data.map(item => new Manga(this, provider, item.hid, item.title.trim()));
         } catch { // TODO: Do not return empty list for generic errors
             return [];


MikeZeDev:
49 was chosen when the website was using 49 , it hasn't changed since. It shouldn't "MISS". I mean, on every website, "limit" pick a number of manga in a predetermined list. If you chose 49, and 49 again, then you should get 49 starting for where previous 49 ended.

If what you say is true, Comick uses a twisted logic and in doubt its the case.

"Missing mangas" is just because servers limits api results after a while thats all.

About the other parameter i am conflicted. I mean, its fine but if we start going this way, people may disagree with that choice and complain, want another sorting, make another PR, yada yada.

Arch-Storm:
I think you kinda misunderstood what I meant. Using limit=50 instead of 49 will simply allow to fetch one more manga per get request, which will increase the total fetched from 2450 to 2500. There is zero downside to this from what i can tell.

Haruneko "missing" some of the popular comics is a completely unrelated problem, which happens because the api by default (if no sort parameter is given), just sorts by latest upload date. I guess there is room to discuss what sorting makes the most amount of sense but i feel like it certainly isn't in the users interest to hide the literal most popular comic from them. Hiding something with 174.000 followers to show something with 20 followers instead seems counterproductive for the "average" user. The best thing would be to somehow devise a solution that can fetch *all* comics from the site, for example by fetching comics by release year. there shouldn't be any single year with over 2500 comics. So I could make a for loop to go through all the years one by one. 
Let me know if you'd be happy with that or if you have another idea to fetch more than the 2500 limit.

PR #1092, opened 2025-04-07
Hitomi : add rate limit for pictures

MikeZeDev:
it seems enough to prevent errors 503


web/src/engine/websites/Hitomi.ts
```diff
@@ -18,6 +19,7 @@ export default class extends DecoratableMangaScraper {
 
     public constructor() {
         super('hitomi', `Hitomi`, 'https://hitomi.la', Tags.Media.Manga, Tags.Language.Multilingual, Tags.Rating.Pornographic, Tags.Source.Aggregator, Tags.Accessibility.RegionLocked);
+        this.imageTaskPool.RateLimit = new RateLimit(2, 0.5);
```
ronny1982, line 22: I don't think they are using such a small interval, so `4, 1` should be fine

ronny1982 (approved):

PR #1114, opened 2025-04-15
Add ComiciViewer (template only)

MikeZeDev:
For websites samples check here https://github.com/manga-download/haruneko/pull/507


ronny1982 (approved):
Approved without review

PR #1115, opened 2025-04-16
testing mixed content workaround

MikeZeDev:
perhaps we could use 'Upgrade-Insecure-Requests' header in pictures requests instead?

### DIFF
--- a/web/src/engine/platform/nw/FetchProvider.ts
+++ b/web/src/engine/platform/nw/FetchProvider.ts
@@ -130,6 +130,16 @@ export default class extends FetchProvider {
     public async Fetch(request: Request): Promise<Response> {
         // Fetch API defaults => https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
         await UpdateCookieHeader(request.url, request.headers);
+
+        if (/^http:\/\//.test(request.url)) {
+            const newHeaders = new Headers(request.headers);
+            newHeaders.set('Upgrade-Insecure-Requests', '1');
+            request = new Request(request, {
+                redirect: 'follow',
+                headers: newHeaders
+            });
+        }
+
         const response = await fetch(request);
         await super.ValidateResponse(response);
         return response;


ronny1982 (changes requested):
This seems to be a bad idea to me. The main purpose of our HTML code is to act as UI. Now we are treating the application more like a website with its own origin and leveraging HTML directives to implicitly control the behavior of requests such as i-frames, images and the Fetch API.

The goal would be to perform such requests exactly as they were made from the manga website itself. Unfortunately we making them from the HakuNeko domain instead (therefore the platform specific origin and header shenanigans in our fetch wrapper).

I would suggest to solve this directly in the fetch wrapper or (if it can't be avoided) with Electron/NW.js platform specific solutions.

MikeZeDev:
yeah id rather change requests too considering its what real browsers does. 
it was just a poc. gotta work on that insecure upgrade header tomorrow.

MikeZeDev:
Meh. using that flag in requests is pointless.

```ts
    public async Fetch(request: Request): Promise<Response> {
        // Fetch API defaults => https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
        await UpdateCookieHeader(request.url, request.headers);

        if (/^http:\/\//.test(request.url)) {
            const newHeaders = new Headers(request.headers);
            newHeaders.set('Upgrade-Insecure-Requests', '1');
            request = new Request(request, {
                redirect: 'follow',
                headers: newHeaders
            });
        }

        const response = await fetch(request);
        await super.ValidateResponse(response);
        return response;
    }
}
```

Because the fetch is ultimately blocked by chrome engine for mixed content.

Thats unfortunate. Launching the binary using  `--allow-running-insecure-content` allow the request to be performed and the server ITSELF automatically upgrade http link to https (without any code change).

Since changing the CSP of the main page is out bad, and using that command line switch is out of the question, the solution would be 

* changing http to https on per website basis.
* changing http to https on Fetch wrapper (which can cause unwanted effects)

Closing this. Experiments over.

PR #1120, opened 2025-04-19
Add MangaPark

MikeZeDev:
We had 2 Mangapark but the second one is a mirror of the first. One of many.
About languages tags : mangas in their database only use those two languages.


ronny1982 (approved):
Approved without review

PR #1137, opened 2025-05-01
Startup guide

Sheepux:
<WIP, desployed for UI review>
wont be merged without translation


MikeZeDev:
Thats very nice. However, i shouldn't be able to click on "Complete" or next steps unless i setup a manga folder. I clicked on it but canceled :)

PR #1148, opened 2025-05-08
Add MeianPlus

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/7891

Filtering non ebooks series, non ebook "chapters" (volumes)


web/src/engine/websites/MeianPlus.ts
```diff
@@ -0,0 +1,160 @@
...
+    }[]
+}
+
+type BlockInfo = {
+    w: number,
+    h: number
+}
+
+const tokenScript = `decodeURIComponent(document.cookie.match(/token_meian_plus=([^;]+)/).at(1)).replaceAll('"', '');`;
+const scramblingMatrix: Array<Array<number>> = [];
```
ronny1982, line 53: Make this a class member
Also use map
```javascript
const scramblingMatrix = new Array(100).fill(null).map((_, index) => [ ((index / 10) >> 0) + 1, ((index % 10) >> 0) + 1 ]);
```

web/src/engine/websites/MeianPlus.ts
```diff
@@ -0,0 +1,160 @@
...
+        param: string
+    }[]
+}
+
+type BlockInfo = {
+    w: number,
+    h: number
+}
+
+const tokenScript = `decodeURIComponent(document.cookie.match(/token_meian_plus=([^;]+)/).at(1)).replaceAll('"', '');`;
```
ronny1982, line 52: Maybe this can be simplified
`cookieStore.get('token_meian_plus')).then(cookie => cookie.value);`

web/src/engine/websites/MeianPlus.ts
```diff
@@ -0,0 +1,160 @@
...
+        this.token = await FetchWindowScript(new Request(this.URI), tokenScript, 2500);
+    }
+
+    public override ValidateMangaURL(url: string): boolean {
+        return new RegExpSafe(`^${this.URI.origin}/catalogue/licence/[^/]+/\\d+$`).test(url);
+    }
+
+    public override async FetchManga(provider: MangaPlugin, url: string): Promise<Manga> {
+        const idLicence = url.match(/\/licence\/[^/]+\/(\d+)$/).at(1);
+        const { licence: { id_licence, titre_licence } } = await this.FetchJSON<APISerie>(this.CreateRequest(`./licence/?id_licence=${idLicence}`));
```
ronny1982, line 84: Apply DRY principle, introduce a method to `FetchAPI` and create the request inside this method
Update `this.token` from within this method (only when necessary)
Don't use `this.token` outside the `FetchAPI` method

ronny1982:
Wait until refactoring of `MangaLib` is finished and use a similar pattern for request and token handling

ronny1982 (changes requested):

web/src/engine/websites/MeianPlus.ts
MikeZeDev: That website is somehow proxying cookiestore and its ugly to work with ; ofc i would have used cookiestore instead.

web/src/engine/websites/MeianPlus.ts
MikeZeDev: > Update this.token from within this method (only when necessary)

should i use base url provides for checking? As its only necessary when gathering pages.

web/src/engine/websites/MeianPlus.ts
ronny1982: So filtering all cookies does also not work?
`cookieStore.getAll().then(cookies => cookies.find(cookie => cookie.name === 'token_meian_plus')?.value);`

ronny1982: The `cookieStore` works for me, when evaluating the promise.

web/src/engine/websites/MeianPlus.ts
```diff
@@ -0,0 +1,183 @@
...
+    public override async FetchPages(chapter: Chapter): Promise<Page<BlockInfo>[]> {
+        //get pages number
+        const { success, ebook } = await this.FetchAPI<APIEbook>(`./ebook/?ref=${chapter.Identifier}`);
+
+        if (!success) {
+            throw new Exception(R.Plugin_Common_Chapter_UnavailableError);
+        }
+
+        //get pagesinfo
+        const { images } = await this.FetchAPI<APIImages>(`./v1/images/?ref=${chapter.Identifier}&p=1&w=1920&h=1080&q=1&webp=true=&nb_pages=${ebook.max_page}&devicePixelRatio=1.5`, this.imageCDN);
```
ronny1982, line 148: Can there be multiple pages?

ronny1982, line 148: Is a higher resolution other than 1920x1080 supported?

ronny1982, line 148: Is the device ratio required? (also desktop screens are usually 16/9)

web/src/engine/websites/MeianPlus.ts
```diff
@@ -0,0 +1,183 @@
...
+    #token: string = null;
+
+    constructor(private readonly clientURI: URL) { }
+
+    /**
+     * Extract the token directly from the website (e.g., after login/logout through manual website interaction)
+     */
+    public async UpdateToken() {
+        try {
+            this.#token = await FetchWindowScript<string>(new Request(this.clientURI), `(async () => (decodeURIComponent((await cookieStore.get('token_meian_plus')).value).replaceAll('"', '')))();`) ?? null;
```
ronny1982, line 66: An async function always return a `Promise` so it can never be `null`, therefore `?? null` has no effect

ronny1982: In this special case I would prefer `JSON.parse` to get the string from the enclosed double quotes `"` instead of `replaceAll` so the intention is more clear (`replaceAll` could be confused with replacing more than the enclosing double quotes)

ronny1982: To avoid script errors (handle `null`) this could be re-arranged to
```javascript
this.#token = JSON.parse(decodeURIComponent(await FetchWindowScript<string>(new Request(this.clientURI), `(async () => (await cookieStore.get('token_meian_plus')).value)();`) ?? null));
```

Issue #1159, opened 2025-05-15
No longer download from MagazinePocket


ancoromotchy:
The system on Magazine Pocket (https://pocket.shonenmagazine.com/) seems to have changed, the list of comics does not load correctly and cannot be downloaded.

ruk011:
How do install the update?
I can't find the folder.

MikeZeDev:
you dont, you wait

PR #1162, opened 2025-05-16
ShonenMagazine : rewrite website

MikeZeDev:
its now very similar to CiaoPlus

Closes https://github.com/manga-download/haruneko/issues/1159

Ref: https://github.com/manga-download/hakuneko/issues/7930


ronny1982 (approved):
Approved without review

PR #1176, opened 2025-05-27
ColaManga: fix for unencrypted pictures

MikeZeDev:


ronny1982:
> With your [suggestion](https://github.com/manga-download/haruneko-proprietary/blob/a3e4a1f2eca3310040afa8e0feeb5939c5d02099/haruneko/web/src/engine/websites/ColaManga.ts#L27), promise would never be resolved on chapters with non encrypted images. That and multiple problems, for example you could reject too early when the next call would work

Removed my comments and suggestions due to a refactoring of _ColaManga_ based on the contribution of this PR

PR #1181, opened 2025-05-31
CoreView : add volume & magazine support

MikeZeDev:
* id doesnt change
* we now try to fetch volumes along with chapters
* Fetching all magazines too when possible.
* Comic Growl and Shonen Magazine dont use Coreview anymore


web/src/engine/websites/decorators/CoreView.ts
```diff
@@ -100,36 +103,36 @@ export function ChaptersSinglePageAJAXV1(query: string = defaultQueryChapters, e
...
+        try {
+            const { html, nextUrl } = await FetchJSON<APIChaptersHTML>(new Request(endpoint));
+            const doc = new DOMParser().parseFromString(html, 'text/html');
+            chapters.push(...[...doc.querySelectorAll<HTMLAnchorElement>(queryChapters)].map(chapter => {
+                const { id, title } = extractor(chapter);
+                return new Chapter(this, manga, id, title.replace(manga.Title, '').trim() || title);
+            }));
+            endpoint = nextUrl;
+        } catch {
+            run = false;
```
27136773: With the Pokemon exception handling, any error (network connection lost, out of system memory, ...) will result in a valid chapter list, even when incomplete.

Should clearly define the condition when run is set to `false`

web/src/engine/websites/ComicDays_e2e.ts
```diff
@@ -19,4 +20,48 @@ new TestFixture({
...
+new TestFixture({
+    plugin: {
+        id: 'comicdays',
+        title: 'コミックDAYS (Comic Days)'
+    },
+    container: {
+        url: 'https://comic-days.com/volume/2550689798642753315',
+        id: '/volume/2550689798642753315',
+        title: 'ＪＪＭ 女子柔道部物語 社会人編',
+    }, /* Paid content
```
27136773, line 35: Arn't there no free chapters?

web/src/engine/websites/decorators/CoreView.ts
MikeZeDev: Well, not my fault the api always specify a nextUrl and it always end with network error. There is nothing to see, nothing to check. Unless i am wrong and that "nexUrl" is here for decoration and all "chapters" are always contained in first page.

Its for magazines its not easy to test, its always paid content.

27136773: What is the `nextUrl` does it repeat after all chapters are acquired (stays the same)?
How about using `isMissingLastItemFrom`?

web/src/engine/websites/ComicDays_e2e.ts
MikeZeDev: No free **volumes**, no. No website gives entire free volumes to test. They sell volumes after all.

web/src/engine/websites/decorators/CoreView.ts
MikeZeDev: To be honest i am not really sure about what to do.

It seems first request already gather all other magazines from the same magazine, as any request involving nextUrl ends up in failure.

With the previous "V1" method we were building the request ourselves using arbitrary values as parameters (minimum and large one) but it turns our those parameters are sometimes dates, or weird numbers.

So i ended up getting the url from the API..

Testing will require someone that pay from a couple of magazine perhaps?

MikeZeDev: Revised extraction with latest commit.

web/src/engine/websites/decorators/CoreView.ts
```diff
@@ -97,70 +91,90 @@ export function ChaptersSinglePageAJAXV1(query: string = defaultQueryChapters, e
...
+ * @param count - How many pages to fetch, because the api ALWAYS return a "nextUrl" so its unreliable.
+ * @param queryChapters - A CSS selector used to extract chapters from the api answer (html from json)
+ * @param extractor - A function to extract id and title from queried elements
+ */
+async function AjaxFetchEntriesFromHTML(this: MangaScraper, manga: Manga, endpoint: string, count: number = 1, queryChapters: string = defaultQueryChapters, extractor = ChapterExtractor): Promise<Chapter[]> {
+    const chaptersList: Chapter[] = [];
+    for (let i = 1; i <= count; i++) {
+        const { html, nextUrl } = await FetchJSON<APIChaptersHTML>(new Request(new URL(endpoint, this.URI)));
+        const doc = new DOMParser().parseFromString(html, 'text/html');
+        chaptersList.push(...[...doc.querySelectorAll<HTMLAnchorElement>(queryChapters)].map(chapter => {
```
27136773: Consider using `Nodes.entries().map()` to avoid needless spreading

Issue #1183, opened 2025-06-01
Update for Raijinscans website - new look


BaptisteP06:
Hello, the RaijinScan site has changed its look, but it's still under the Madara theme, I think.
( https://raijinscan.fr/ )

MikeZeDev:
![Image](https://github.com/user-attachments/assets/f930ff33-8d40-4669-a973-44c2eda7d815)

BaptisteP06:
Hello, is the problem solved or in the process of being solved? I always get an error when I try to retrieve chapters.

Thanks

MikeZeDev:
Please no non-constructive spam asking for ETA.

MikeZeDev:
Website seems to be fighting vs Mihon, or us, or both. Expect further breaking. In that case open another issue.

PR #1184, opened 2025-06-01
RaijinScans : fix getting page links

MikeZeDev:
+ madara fix to get postid in clipboard method. I have seen this on many websites i think its worth including now.

Closes https://github.com/manga-download/haruneko/issues/1183


MikeZeDev:
I'd better check madara websites now :D

BaptisteP06:
I can't see the chapters at all now xD, it feels like infinite loading or is it my hakuneko bug.

image

PR #1214, opened 2025-06-26
ShonenMagazine: Fix the dimensions of the pieces

clfws9:
Depending on the dimensions of the image, it may not be correctly unscrambled.
This issue can occur if either the width or height of the image is not an integer multiple of 32.

In that case, the dimensions of each piece into which the image is divided will be incorrect.
It seems that the range into which the image is divided must be a multiple of 32.
This method of division has not changed since the website was redesigned.
https://github.com/manga-download/hakuneko/blob/3b87885a8af64d26f9b2fa79c2ad8754b3ecf764/src/web/mjs/connectors/templates/CoreView.mjs#L130-L133

### To Reproduce

1. Name of connector: `週刊少年マガジ (Weekly Shonen Magazine & Pocket Magazine)`
2. Name of manga: `頭文字D`
3. Name of chapter: `【Vol.1】ハチロク買おーぜ`
   - URL of chapter: https://pocket.shonenmagazine.com/title/00553/episode/209248
4. Compare the downloaded image with the one displayed on the website.

#### Expected behavior

The downloaded image matches the one displayed on the website.
Also, the width of the piece is 168 px.

#### Actual behavior

The downloaded image does not match the one displayed on the website.
Also, the width of the piece is **170** px.

### DIFF
--- a/web/src/engine/websites/CiaoPlus.ts
+++ b/web/src/engine/websites/CiaoPlus.ts
@@ -49,6 +49,7 @@ export default class extends DecoratableMangaScraper {
     protected apiUrl = 'https://api.ciao.shogakukan.co.jp/';
     protected requestHashProperty = 'x-bambi-hash';
     protected requestHashAppend : string = '';
+    protected pieceScale = 8;
 
     public constructor(id = 'ciaoplus', label = 'Ciao Plus', url = 'https://ciao.shogakukan.co.jp', tags = [Tags.Media.Manga, Tags.Language.Japanese, Tags.Source.Aggregator]) {
         super(id, label, url, ...tags );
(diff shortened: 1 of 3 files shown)


MikeZeDev:
Nice catch.

>  This method of division has not changed since the website was redesigned.

Yeah, but its does not seems to be Coreview anymore so that's irrelevant.

First i made CiaoPlus i merely pasted their own code.
Then i saw Pocket changed and it was the same system. The manga in PocketMagazine tests was fine and still , same for other i tested 🤷 . Ah, well, that slipped past me.

MikeZeDev:
It seems the difference is not 8 or 32. After comparing both websites, they all use 8 but code used to compute piece size is slightly different. I am gonna dig and perhaps fix it myself. I'd rather override a function than rely on an arbitrary value :)

PR #1288, opened 2025-08-08
WordPressMadara : remove manga title from chapter title

MikeZeDev:
I think it should work like Common.
1) Fetch
2) call extractor. Extractor has no access to manga.title
3) replace "manga.title" on extractor result
Common.FetchSinglePageMangaCSS does that. 
WordPressMadara FetchChaptersCSS omit the last part.
It would prevent some dirty hack like on https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/TortugaCeviri.ts#L21
Ofc for ToonFR we would still need to replace ... remaining '- ' 😄 
Or https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/ManhwaLatino.ts


MikeZeDev:
Changes done separatly. Closing.

Issue #1311, opened 2025-08-19
What about refactoring the entire project in HaruNeko


Zephyris-Pro:
While looking at the code, I noticed that some parts, such as dependencies, builds, and other elements, still use the name HakuNeko. Is there a particular reason for keeping this name (such as translation keys or something else)? Wouldn't it be better to refactor all these names for consistency?

MikeZeDev:
* Connectors (websites) id are intentionally kept the same, to facilitate bookmarks import from Hakuneko.

* There was no translation in Hakuneko

* As for dependencies well, it...depends. We evaluated some libraries , kept some, removed others. I took on me to remove CryptoJS when we could as Subtle is more than enough, stuff like that. Another pdf library we tried had trouble to recognize certain (perfectly valid) JPEG files, etc... Cant really talk much.

* We are constantly refactoring websites plugins to our latest "standards" when we have the opportunity to touch a website code again. It means yeah some may need revisions. 

* I cant talk much about the engine.

* Its meant to REPLACE Hakuneko anyway

If you have suggestions or examples, please go ahead, show us. I may talk a lot but i am not the boss.

Sheepux:
haruneko is just the codename for hakuneko v10
Haruneko will be merged into hakuneko's repository once it's feature-complete (iso with haku), for now "haru" is just a codename.

PR #1312, opened 2025-08-19
Refactoring HakuNeko -> HaruNeko

Zephyris-Pro:
As I mentioned here: https://github.com/manga-download/haruneko/issues/1311, some refactoring was needed for consistency.

I’ve updated all mentions of **HakuNeko** to **HaruNeko** manually.
I left the string **"Import HakuNeko's bookmarks from previous version"**, since users may still want to migrate their bookmarks from HakuNeko.

I’ve tested the app with these changes, and it is working correctly.
If any additional changes are needed, please let me know and I’ll be happy to update.


MikeZeDev:
Whats the reasoning behind this?

Zephyris-Pro:
As you mentioned earlier, HaruNeko is supposed to replace HakuNeko. However, there are still many references to HakuNeko in the codebase, which gives the impression that a lot of code was simply copied over (which is probably true in some cases and that's fine, better than rewriting everything from scratch).

The problem is that for someone who isn't already familiar with the code, this is confusing. The project is called HaruNeko, yet references to HakuNeko keep appearing everywhere, even though HaruNeko is meant to be the successor. This inconsistency makes it harder to understand for new contributors .

I think the code should be refactored to consistently use HaruNeko and avoid residual references to HakuNeko. That way, contributors and new developers won’t get lost, and the project will feel more coherent and professional.

Sheepux:
There are no plans to let haruneko as the final name. https://github.com/manga-download/haruneko/issues/1311#issuecomment-3207350036

MikeZeDev:
As Sheep said, pretty sure that PR is pointless.

27136773:
Closed: At some point the name will be changed to `HakuNeko`, which contradicts the purpose of this PR

PR #1328, opened 2025-08-29
Vtheme : improve pages sorting

MikeZeDev:
Depending on a server flag, website could use another sorting method. Was found on at least 2 websites (albeit not always activated)

Reported on Discord

### DIFF
--- a/web/src/engine/websites/templates/VTheme.ts
+++ b/web/src/engine/websites/templates/VTheme.ts
@@ -12,7 +12,7 @@ type APIManga = {
 };
 
 type APIMangas = {
-    posts: APIManga[ 'post' ][];
+    posts: APIManga['post'][];
 };
 
 type APIChapters = {
@@ -31,12 +31,14 @@ type APIChapters = {
 
 type HydratedPages = {
     images: {
-        url: string
+        url: string;
+        order: number;
     }[]
 };
 
 @Common.ImageAjax()
 export class VTheme extends DecoratableMangaScraper {
+    protected useAlternativeSorting: boolean = false;
 
     private readonly apiUrl = (() => {
         const uri = new URL(this.URI);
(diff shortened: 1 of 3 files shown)


web/src/engine/websites/templates/VTheme.ts
```diff
@@ -87,8 +89,23 @@ export class VTheme extends DecoratableMangaScraper {
             });
     }
 
+    private ToNumber(text: string): number {
+        return /(\d+)/.test(text) ? parseInt(text.match(/(\d+)/).at(1)) : Number.POSITIVE_INFINITY;
+    }
+
     public override async FetchPages(chapter: Chapter): Promise<Page[]> {
         const { images } = await FetchNextJS<HydratedPages>(new Request(new URL(chapter.Identifier, this.URI)), data => 'images' in data);
```
27136773: Consider using FetchCSS or FetchJS instead
`[ ...document.querySelectorAll('.image-container img[data-image-index]') ].map(img => img.src)`

web/src/engine/websites/templates/VTheme.ts
```diff
@@ -87,8 +89,23 @@ export class VTheme extends DecoratableMangaScraper {
...
+        if (this.useAlternativeSorting) {
```
27136773: Consider to always apply this sorting instead of introducing an additional flag.

web/src/engine/websites/templates/VTheme.ts
27136773: e.g., for _VTheme_ decorator
```typescript
@Common.PagesSinglePageCSS('.image-container img[data-image-index]')
```

e.g., for _VortexScans_ decorator
```typescript
@Common.PagesSinglePageJS(`
    new Promise(resolve => {
        const images = [ ...document.querySelectorAll('.image-container img[data-image-index]') ];
        images.forEach(img => img.scrollIntoView());
        setTimeout(() => resolve(images.map(img => img.src)), 500);
    });
`)
```

MikeZeDev: I dont really like that method of waiting for images, i feel its not safe because if rely in the user connection speed but okay.

web/src/engine/websites/templates/VTheme.ts
MikeZeDev: Every websites sort first according to `order`. For two of them there is that custom sorting right **after** : Hivetoon got it **disabled** (its a env flag), on Magustoon its **forced**.

I'd rather just include mandatory sorting in decorator, then override `FetchChapters` on Magustoon to sort pages like the websites wants. I feel its safer to follow the code than just "custom sort will work everywhere".
I have a feeling it may move some pages to the wrong place.

```ts
const pages = await super.Fetchapters.call(this, chapter);
return pages.sort( ....custom magus sorting)
```

MikeZeDev: Ofc if using CSS or JS we dont have to care about that

web/src/engine/websites/templates/VTheme.ts
MikeZeDev: Script doesnt work btw. Even adjusting delay doesnt work. I dont get it
image

NyxScans is affected too.

27136773: The script is only for VortexScan

MikeZeDev: It doesnt change the fact that is not working for Vortex and i cant find why.

MikeZeDev:
Moved QuantumScans to Vtheme and made adjustements.

PR #1329, opened 2025-08-30
Add KimiComi (Comici)

WhonderWy:
Attempt to add KimiComi


MikeZeDev:
Well.. https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/KimiComi.ts

PR #1361, opened 2025-09-22
Ganma : rewrite website

MikeZeDev:


web/src/engine/websites/Ganma.ts
```diff
@@ -50,72 +70,88 @@ export default class extends DecoratableMangaScraper {
...
+        return new Array(pageCount).fill(0).map((_, index) => index + 1).map(page => new Page(this, chapter, new URL(`${pageImageBaseURL}${page}.jpg?${pageImageSign}`)));
+    }
+
+    private async FetchAPI<T extends JSONElement>(operation: string, variables: JSONObject, useAppHeaders: boolean = true): Promise<T> {
+        const request = new Request(new URL(this.apiUrl), {
             headers: {
-                'X-From': this.URI.href
+                'X-From': this.URI.href + 'web',
+                'Content-Type': 'application/json;charset=UTF-8',
+                Accept: 'application/json, text/plain, */*'
```
27136773: Is this really required?
Also I think when using GraphQL, `json` is the only accepted format.`

web/src/engine/websites/Ganma.ts
```diff
@@ -2,45 +2,65 @@
...
+type APIMagazine = {
+    magazine: APIMagazineDetails;
+};
+
+type PageInfo = {
+    endCursor: string;
+    hasNextPage: boolean;
+};
+
+type APIMagazineDetails = {
```
27136773, line 29: Are all these fields really used?
Remove unused fields.

web/src/engine/websites/Ganma.ts
```diff
@@ -50,72 +70,88 @@ export default class extends DecoratableMangaScraper {
         return icon;
     }
 
+    public override async Initialize(): Promise<void> {
+        super.Initialize();
+        //Website forces persistedQueries so we gather queries hashes from scripts
```
27136773: When the queries are persisted, aren't they supposed to be reliable constant?
Just use the constants for the few operations that we need directly instead of acquire all of them during runtime.

web/src/engine/websites/Ganma.ts
```diff
@@ -50,72 +70,88 @@ export default class extends DecoratableMangaScraper {
...
-                'X-From': this.URI.href
+                'X-From': this.URI.href + 'web',
+                'Content-Type': 'application/json;charset=UTF-8',
+                Accept: 'application/json, text/plain, */*'
+            }
+        });
+        if (useAppHeaders) request.headers.set('User-Agent', 'GanmaReader / 9.9.1 Android');
+        return FetchGraphQL<T>(request, operation, undefined, variables, {
+            persistedQuery: {
+                sha256Hash: this.queryHashesMap.get(operation),
```
27136773: Consider providing the hash for the persisted query as parameter to the enclosing function (see also Senkuro.ts)

web/src/engine/websites/Ganma.ts
MikeZeDev: That website moved from queries to persistedQueries even before the previous PR was done.
Maybe its intentional, or not. 

What it i keep that acquiring but i set default values as a fallback?

web/src/engine/websites/Ganma.ts
MikeZeDev: They are all used but ill double check.

endCursor is needed but its always present even on last page so hasNextPage is needed too.
isWebOnlySensitive is a flag for the few manga that are web only and it guarantee we use the good api parameters.

web/src/engine/websites/Ganma.ts
MikeZeDev: ok... can i keep the map for readability ?

PR #1406, opened 2025-11-02
Fix MangaCrab not working

Zephyris-Pro:
`PagesSinglePageCSS` has now been replaced by `PagesSinglePageJS`.
 I'm not sure if a timeout of 0 seconds is appropriate.
 The `data-img-*` attribute has been changed to `wp-manga-chapter-img`, and the URL is now retrieved using the `zdk2-*` attribute.

### DIFF
--- a/web/src/engine/websites/MangaCrab.ts
+++ b/web/src/engine/websites/MangaCrab.ts
@@ -12,10 +12,8 @@ const chapterScript = `
 
 const pageScript = `
     [...document.querySelectorAll('div.page-break img.wp-manga-chapter-img[id^="image-"]')]
-    .flatMap(img => {
-        const k = img.getAttributeNames().find(n => n.startsWith('zdk2-'));
-        return k ? [img.getAttribute(k)] : [];
-    });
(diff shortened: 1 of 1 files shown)


web/src/engine/websites/MangaCrab.ts
```diff
@@ -10,10 +10,18 @@ const chapterScript = `
     }));
 `;
 
+const pageScript = `
+    [...document.querySelectorAll('div.page-break img.wp-manga-chapter-img[id^="image-"]')]
```
27136773, line 14: Suggestion to use URL pattern matching of attribute values instead of possible rotating attribute name

```js
[...document.querySelectorAll('div.page-break img.wp-manga-chapter-img[id^="image-"]')]
    .map(img => [...img.attributes].find(attribute => attribute.value.startsWith('/validate2.php'))?.value)
    .filter(img => img);
```

Zephyris-Pro: Yup good idea !

manga-download-owner (approved):

27136773 (approved):

PR #1418, opened 2025-11-16
WordPressMadara : add e2e file builder

ghost:


27136773:
Closed, I don't see a benefit to add this into the repository. Single developer may keep the script as they are pleased.

PR #1454, opened 2025-11-28
feat:Port MyAnimeListManga From Hakuneko

ghost:


web/src/engine/websites/MyAnimeListManga.ts
```diff
@@ -0,0 +1,93 @@
...
+        return filenames.map(file => {
+            const uri = new URL(image_base_url + '/' + file);
+            uri.search = query_params_part;
+            return new Page<PageMode>(this, chapter, uri, { mode: 'xor' });
+        });
+    }
+
+    public override async FetchImage(page: Page<PageMode>, priority: Priority, signal: AbortSignal): Promise<Blob> {
+        const blob = await Common.FetchImageAjax.call(this, page, priority, signal);
+        switch (page.Parameters.mode) {
```
27136773: Missing case, what when `mode` is not `xor`?

web/src/engine/websites/MyAnimeListManga.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const [button] = await FetchCSS<HTMLDivElement>(new Request(new URL(manga.Identifier, this.URI)), 'div.v-manga-store-purchase-bulk-button');
+        const chaptersData: ChaptersJSON = JSON.parse(button.dataset['items']);
+        return chaptersData.filter(({ is_free: free, is_possessed: owned, is_previewable: preview }) => preview || free || owned)
```
27136773, line 50: Not sure if providing previews make sense.
Users may complain about incomplete chapters.

web/src/engine/websites/MyAnimeListManga.ts
```diff
@@ -0,0 +1,93 @@
...
+                const isPossessed = owned ?? false;
+                const isFull = free || isPossessed;
+                const title = [name.trim(), isFull ? '' : '[Preview]'].join(' ').trim();
+                return new Chapter(this, manga, new URL(isFull ? vUrl : pUrl, this.URI).pathname, title);
+            });
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page<PageMode>[]> {
+        const [viewerElement] = await FetchCSS<HTMLDivElement>(new Request(new URL(chapter.Identifier, this.URI)), 'div.v-manga-store-viewer');
+        if (!viewerElement) {
```
27136773, line 66: Is it safe to assume that the user needs to login or purchase the chapter because the element is missing?
Can the element be missing because of other reasons?

web/src/engine/websites/MyAnimeListManga.ts
```diff
@@ -0,0 +1,93 @@
...
+            throw new Exception(R.Plugin_Common_Chapter_UnavailableError);
+        }
+        const { manuscript: { filenames, image_base_url, query_params_part } }: ViewerJSON = JSON.parse(viewerElement.dataset.state);
+        return filenames.map(file => {
+            const uri = new URL(image_base_url + '/' + file);
+            uri.search = query_params_part;
+            return new Page<PageMode>(this, chapter, uri, { mode: 'xor' });
```
27136773: When there is only a single option, then no need to add this option at all.

27136773:
Cannot verify that this is really working due to lack of free chapters and account.

web/src/engine/websites/MyAnimeListManga.ts
MikeZeDev: the other case is Publus but we cant handle it yet.

In that case we wont be here and be stopped way before.

web/src/engine/websites/MyAnimeListManga.ts
MikeZeDev: * People may want to see preview as well
* Other websites also got previews and we are fine with it until now.

web/src/engine/websites/MyAnimeListManga.ts
MikeZeDev: Like ive said it could be missing for that reason OR because its PUBLUS (novels)

And there is no chapter listed in case its a novel i made sure of it in FetchChapters.

web/src/engine/websites/MyAnimeListManga.ts
MikeZeDev: For NOW. Its provisional as we cant handle Publus yet

MikeZeDev:
So its also providing NOVELS and novels use PUBLUS : 

https://myanimelist.net/store/manga_volume/9654/Spice_and_Wolf_light_novel_Volume_23

We just  cant handle them for now.

web/src/engine/websites/MyAnimeListManga.ts
27136773: In that case the switch is misleading an an `if` should be used

Issue #1450, opened 2025-11-26
Nyx Scans:  Chapter download will not start


Tony763:
Hello, with Nyx Scans, media / chapters are listed and can be added to download list but download will not start. (free chapters)

For example: https://nyxscans.com/series/ugly-complex

<img width="470" height="445" alt="Image" src="https://github.com/user-attachments/assets/74658510-817b-4846-a7ea-26774710f812" />

When I add another chapter from other working source, it's download will not start, too.

<img width="470" height="445" alt="Image" src="https://github.com/user-attachments/assets/ba990d94-38c1-42ee-9b85-bcff748a45e8" />

ghost:
Meh. The lazy loading of VTHEME websites (like Nyx) is so annoying...

No problem on rolia tho

<img width="771" height="396" alt="Image" src="https://github.com/user-attachments/assets/5ed2b294-b981-4ecd-a839-7554d8e17173" />

ghost:
Could be related https://github.com/manga-download/haruneko/issues/1356

27136773:
I cannot reproduce this issue as of 2026-01-07 with Electron 39.1.0 on MacOS

Tony763:
I am on 06b6dd46b, issue still ocure (linux, electron)

27136773:
A fix has been added to addresses the issue: e2a03358

Tony763:
Tested, works good. Thank You

Issue #1356, opened 2025-09-20
Images not loading on Vortex Scans connector


Zoroam:
Manga images are not loading on vortex scans connector. Website url: https://vortexscans.org/ 

Just 2 images are loading for every chapter

Zephyris-Pro:
They added a captcha-like verification: "Completing challenge ..."

Zephyris-Pro:
Same here: https://github.com/manga-download/haruneko/issues/1450

Zephyris-Pro:
EDIT: it's vShield

MikeZeDev:
Yeah i saw that. It should be trivial to bypass but for some reasons i have trouble to execute script on the website

27136773:
I cannot reproduce this issue as of 2026-01-08 with Electron 39.1.0 on MacOS

MikeZeDev:
Still present as of 2026-01-08 with Electron 39.2.7, 39.1.0 on Windows.
Perhaps its a difference with Mac Binaries?

27136773:
A fix has been added to addresses the issue: e2a03358

MikeZeDev:
Fixe works for me 🆗

PR #1483, opened 2025-12-20
VTHEME : handle lazy loading

MikeZeDev:
This fixes VTHEME lazy loading scripts.

Basically, it allows extension maintainers to execute scripts in a visible window, since its impossible to trigger some lazy loading type in invisible windows. 

Closes https://github.com/manga-download/haruneko/issues/1356
Closes https://github.com/manga-download/haruneko/issues/1450


web/src/engine/platform/electron/RemoteBrowserWindow.ts
```diff
@@ -53,6 +53,7 @@ export default class RemoteBrowserWindow implements IRemoteBrowserWindow {
             height: 800,
             center: true,
             webPreferences: {
+                offscreen: !show, //offscreen = true allows some lazyloading script to work, but window is blank
```
27136773: This can easily be used for webesites to detect HakuNeko
`isOffscreenRendering = window.outerHeight - window.innerHeight < 8;`

web/src/engine/websites/templates/VTheme.ts
```diff
@@ -32,11 +32,14 @@ type APIChapters = {
...
-        const images = [ ...document.querySelectorAll('.image-container img[data-image-index]') ];
-        setInterval(() => {
-            images.forEach(img => img.scrollIntoView());
+        const interval = setInterval(() => {
+            const images = [ ...document.querySelectorAll('.image-container img[data-image-index]') ];
+            images.forEach(img => img.scrollIntoView({ block: "center" }));
             const sources = images.map(img => img.src || '');
-            if(!sources.some(src => src === '')) resolve(sources);
+            if (!sources.some(src => src === '')) {
+                clearInterval(interval);
```
27136773: No need to clear the interval, the window will be disposed as soon as the promise is resolved.

web/src/engine/platform/electron/RemoteBrowserWindow.ts
27136773: Should also keep the framerate low to safe resources such as CPU/GPU and Memory
`win.webContents.setFrameRate(5);`

web/src/engine/platform/electron/RemoteBrowserWindow.ts
27136773, line 48: If offscreeen rendering is really required, it should be added as parameter with default set to `false` so only the website that really needs it could set it to `true`.

web/src/engine/websites/NyxScans.ts
```diff
@@ -1,7 +1,13 @@
-import { Tags } from '../Tags';
+﻿import { Tags } from '../Tags';
+import { AddAntiScrapingDetection, FetchRedirection } from '../platform/AntiScrapingDetection';
 import icon from './NyxScans.webp';
 import { VTheme } from './templates/VTheme';
 
+AddAntiScrapingDetection(async (invoke) => {
```
27136773, line 7: This check will be added globally and performed for all websites.
Is this intentional?

MikeZeDev: Like all AddAntiScrapingDetection checks, yes.  Its valid for Nyx and Vortex tho.

web/src/engine/platform/electron/RemoteBrowserWindow.ts
MikeZeDev: Yeah i was thinking about something like that too.  Not currently seeing the code but i wonder if we need to adapt the interface and by doing so ....this parameter would be useless in NWJS ...

web/src/engine/platform/electron/RemoteBrowserWindow.ts
MikeZeDev: Agreed  . We could still get JS workarounds fro this detection i guess.

web/src/engine/websites/NyxScans.ts
```diff
@@ -1,7 +1,13 @@
...
+    const result = await invoke<boolean>(`document.querySelector('body main.card > h1')?.textContent.trim() == 'One more step'`);
```
27136773: Use `===`

web/src/engine/platform/electron/RemoteBrowserWindow.ts
MikeZeDev: The big problem with offscreen is that since website is using turnstile we then show the window but because of offscreen its blank, so its a bad solution :/ 

I mean if we really need a visible window, lets go for it and make it visible but tiny and make sure to resize it to a proper size in case we need to show it for captcha. 

A tiny window could be detected as suspect but 🤷 that would at least cause less to no problems with captchas.

27136773 (commented):
From a users and developers point of view I really don't like this solution of showing a browser window every time images are extracted. I think there should be a better solution (e.g., hooking into observers).

MikeZeDev:
I am not satisfied myself. The problem lies within chrome internal working and i dont think we can hack our way using hooks or monkey patching. However what i know is that id rather use that solution than let users unable to use those websites and raise a flag for websites telling them to use that lazy load method. 

Offscreen works until you need to display the window for anti scraping or captcha . In that case you need to recreate a new window to display anything and thats way more ugly if we consider our workflow imho.

Not to mention offscreen is a electron only thing. 

I dont have a better solution right now  despites my research. Sorry.

Edit none of the proposed scripts worked for me until i made the window visible with a reasonable size .

27136773:
I cannot reproduce the problem with Electron 39.1.0 release on MacOS (residential IP and bad reputation IP) for
- Hijala Scans
- MagusManga
- Nyx Scans
- Vortex Scans

All these website running fine for me without these changes.

MikeZeDev:
Meh. I encounter the same problem than mentioned in issues but you dont. :/ And in theory you should too .. 🤔 
Windows, Electron, NWJS. Thats disturbing.

27136773:
Background throttling is now disabled: e2a03358

PR #1496, opened 2025-12-27
Desu : fix VS typo

f5xyhmt849ek:
I make fixes based on the developer console. Therefore, I don't know if there are any other bugs in this connector. Because I can't get around the first problem.
In HakuNeko, the connectors are located in the “/cache/mjs/connectors” folder. In HaruNeko, are the connectors packed, or can they also be found somewhere?


MikeZeDev:
It seems Visual Studio code completion put extra spaces without asking..... Meh, sorry about that.
Dont bother touching Hakuneko.

Issue #1504, opened 2026-01-06
New Issue, Can't connect to the web


BeanDau2212:
I searched on the search bar and found the comic page, but when I clicked to download it, the page showed the following error. Please check and fix it as soon as possible. Thank you very much.

<img width="1919" height="1017" alt="Image" src="https://github.com/user-attachments/assets/0211bfbc-1f5d-4584-8796-053ac8a659a8" />

MikeZeDev:
> Please check and fix it as soon as possible

A friendly advice : never ask like that. "As soon as possible" could be badly received by contributors, especially when they do that on their free time, for free. ;)

Then, It would be better to name the problematic website in issue title. Not a big deal tho. Thanks for reporting 👍

MikeZeDev:
Manga list fix is found, turns out they also changed their picture "protection" again.

PR #1510, opened 2026-01-10
fix: Yurigarden : get mangas and pictures

MikeZeDev:
Closes https://github.com/manga-download/haruneko/issues/1504

### DIFF
--- a/web/src/engine/websites/YuriGarden.ts
+++ b/web/src/engine/websites/YuriGarden.ts
@@ -31,42 +31,44 @@ type PageKey = {
     key?: number[];
 };
 
-const pageScript = `
-    new Promise(async resolve => {
-        let exports = undefined;
+function pageScript(apiUrl: string) {
+    return `
+        new Promise(async resolve => {
+            let exports = undefined;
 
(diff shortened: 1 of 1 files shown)


web/src/engine/websites/YuriGarden.ts
```diff
@@ -22,33 +23,47 @@ type APIChapter = {
...
+            WebAssembly.instantiateStreaming = new Proxy(WebAssembly.instantiateStreaming, {
+                async apply(target, thisArg, args) {
+                    const result = await Reflect.apply(target, thisArg, args);
+                    exports = result.instance.exports;
+                    dispatchEvent(new CustomEvent('${ eventName }'));
+                    return result;
+                }
+            });
+
+            addEventListener('${ eventName }', async (_e) => {
```
27136773: I don't see the need of dispatching and listening to a custom event, just fetch the data directly.

MikeZeDev: I need to wait for the wasm to be fully instanciated, but i could use a setInterval instead

PR #1524, opened 2026-01-21
Added 7z Exporter

axeladams:
I like 7z for its more compact compression than zip (LZMA2 vs deflate) so I added a new exporter based on 7z-wasm. The code was tested on both arch linux and windows 11. The automated npm tests also appear to still pass.

Possible issues:
Stores chapter data in a virtual filesystem in memory and might not work on computers with minimal RAM.
Adds a new line that will need to be translated into languages other than English.
Uses C7Z as the typename which might wrongly imply a comic.xml file exists (similar to CBZ).

### DIFF
--- a/web/src/i18n/locales/en_US.ts
+++ b/web/src/i18n/locales/en_US.ts
@@ -167,6 +167,7 @@ const translations: VariantResource = {
   Settings_Global_MangaExportFormatInfo: 'The container format to store the downloaded content for mangas/comics',
   Settings_Global_MangaExportFormat_FolderWithImages: 'Folder with Images',
   Settings_Global_MangaExportFormat_ComicBookArchive: 'Comic Book Archive (*.cbz)',
+  Settings_Global_MangaExportFormat_SevenZipArchive: '7z Archive (*.7z)',
   Settings_Global_MangaExportFormat_ElectronicPublication: 'E-Book Publication (*.epub)',
   Settings_Global_MangaExportFormat_PortableDocumentFormat: 'Portable Document Format (*.pdf)',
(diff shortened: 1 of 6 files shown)


web/src/i18n/locales/en_US.ts
```diff
@@ -167,6 +167,7 @@ const translations: VariantResource = {
   Settings_Global_MangaExportFormatInfo: 'The container format to store the downloaded content for mangas/comics',
   Settings_Global_MangaExportFormat_FolderWithImages: 'Folder with Images',
   Settings_Global_MangaExportFormat_ComicBookArchive: 'Comic Book Archive (*.cbz)',
+  Settings_Global_MangaExportFormat_SevenZipArchive: '7z Archive (*.7z)',
```
MikeZeDev, line 170: Is'nt it missleading? Since you said this PR will generate C7Z but its written *.7z.

axeladams: Sorry. I meant C7Z is the internal variable name I used. The actual user interface and generated archive are *.7z.

27136773:
There is no real-world benefit for C7Z since everything is already covered by CBZ.

axeladams:
Out of curiosity, is there any chance LZMA2 support will be added in the future?

27136773:
The question would be what is the benefit of using LZMA2, since the images are already highly compressed.
What would be the expectation of using LZMA2?

axeladams:
You're right. I just ran a small test on JPGs (both 7z and zip at max compression). 7z is 1% smaller which is probably not worth it to add.

PR #1544, opened 2026-01-31
feat: Add LunarAnimes

MikeZeDev:
https://github.com/manga-download/hakuneko/issues/8383

### DIFF
--- a/web/src/engine/websites/LunarAnimes.ts
+++ b/web/src/engine/websites/LunarAnimes.ts
@@ -26,11 +26,6 @@ type APIPages = {
     };
 };
 
-type ChapterID = {
-    chapterNumber: number;
-    language: string;
-};
-
 const chapterLanguageMap = new Map([
     ['zh', Tags.Language.Chinese],
     ['en', Tags.Language.English],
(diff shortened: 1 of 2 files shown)


web/src/engine/websites/LunarAnimes.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        const { manga } = await FetchJSON<APIMangas>(new Request(new URL('./manga/search?', this.apiUrl)));
+        return manga.map(({ slug, title }) => new Manga(this, provider, slug, title));
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const { data } = await FetchJSON<APIChapters>(new Request(new URL(`./manga/${manga.Identifier}`, this.apiUrl)));
+        return data.map(({ chapter_number: chapterNumber, language }) => {
+            return new Chapter(this, manga, JSON.stringify({ language, chapterNumber }), `Chapter ${ chapterNumber } (${language})`,
```
27136773: Use the path instead of a custom JSON as chapter ID
`./${manga.Identifier}/${chapter_number}?language=${language}`

web/src/engine/websites/LunarAnimes.ts
```diff
@@ -0,0 +1,80 @@
...
+    ['es', Tags.Language.Spanish],
+    ['th', Tags.Language.Thai],
+    ['tr', Tags.Language.Turkish],
+    ['vi', Tags.Language.Vietnamese],
+]);
+
+@Common.MangaCSS<HTMLMetaElement>(/^{origin}\/manga\/[^/]+$/, 'meta[property="og:title"]', (element, uri) => ({ id: uri.pathname.split('/').at(-1), title: element.content.trim() }))
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+    private readonly apiUrl = 'https://api.lunaranime.ru/api/';
```
27136773: All requests are prefixed with `./manga`, consider appending this to the api URL itself

MikeZeDev: Its on ourpose since website provide novels and anime too and i asume prefix would be different but ok.

PR #1552, opened 2026-02-02
fix : remove unicode format characters from filenames

MikeZeDev:
Case : Mangafire 
Manga: Fate/Stay night, heaven feels
Chapter : 40, english

Chapter name got https://unicode-explorer.com/c/200B in its name. This code is accepted by Operating System (Windows 11) but NOT by FileSystem API, causing an error in ImageDirectoryExporter.export.

While we are at it, lets remove all https://www.compart.com/en/unicode/category/Cf. This is what this pr IS DOING

### DIFF
--- a/web/src/engine/StorageController.ts
+++ b/web/src/engine/StorageController.ts
@@ -35,9 +35,13 @@ export function SanitizeFileName(name: string): string {
         '~': '～', //https://unicode-explorer.com/c/FF5E //File System API cannot handle trailing hyphens
     };
 
+    const patternControlCodes = /[\p{Cc}]/gu; // https://en.wikipedia.org/wiki/C0_and_C1_control_codes
+    const patternControlFormat = /[\p{Cf}]/gu; // https://www.compart.com/en/unicode/category/Cf
+
     // TODO: Reserved names? => CON, PRN, AUX, NUL, COM1, LPT1
     return name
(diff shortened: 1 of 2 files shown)


web/src/engine/StorageController.ts
```diff
@@ -37,7 +37,7 @@ export function SanitizeFileName(name: string): string {
 
     // TODO: Reserved names? => CON, PRN, AUX, NUL, COM1, LPT1
     return name
-        .replace(/[\u0000-\u001F\u007F-\u009F]/gu, '') // https://en.wikipedia.org/wiki/C0_and_C1_control_codes
+        .replace(/[\p{Cc}\p{Cf}]/gu, '')// https://en.wikipedia.org/wiki/C0_and_C1_control_codes, https://www.compart.com/en/unicode/category/Cf
```
27136773: Consider introducing named regex patterns to be more descriptive such as `patternInvisibleCharacters = `, `patternControlCharacters = `, ...

web/src/engine/StorageController_test.ts
```diff
@@ -6,7 +6,7 @@ describe('StorageController', () => {
     describe('SanitizeFileName', () => {
 
         it('Should replace forbidden characters', () => {
```
27136773: Use `each` to test certain patterns such as OS forbidden characters, invisible characters and control characters

PR #1589, opened 2026-03-01
fix: Prochan changes

MikeZeDev:
* multipaged manga list (with novels filtering)
* filter novels from clipboard feature
* multipaged chapter list
* handle encrypted puzzle data


MikeZeDev:
Closing : turnstile everywhere.

PR #1601, opened 2026-03-11
increase fluentcore e2e test timeout

MikeZeDev:


MikeZeDev:
60 seconds and  still timeouted? meh

PR #1630, opened 2026-04-05
fix : PlotTwistNoFansub : handle Turnstile

MikeZeDev:
A bit slow, perhaps could be fine fine-tuned.

similar to https://github.com/manga-download/haruneko/pull/1589


MikeZeDev:
Website is down, other does not use turnstile, hum.

MikeZeDev:
Closed. 
Reasons : website is down, new website is similar to MangasNoSekai so another commit is planned.

PR #1662, opened 2026-04-26
feat: Add backend functions & decorators to get real chapters url

MikeZeDev:
As of now, the "Copy URL" right click menu does nothing.

* This PR add the relevant `ChapterURL `decorator in `Common`, that is returning relevant chapter url computer from chapter.
* A new method has been added to `MangaScraper `: `"GetChapterURL(chapter: Chapter): Promise<URL>"`
* Relevant decorator `ChapterURL `calls has been added in templates
* Exceptions (i.e website where chapter id is NOT a pathname) are done. Some website i.e where chapter path are random like `vymanga `were not touched. Other need more work, like proper OPENSSL encryption (yeah, there are website doing that to chapter urls)

* Submitting this as  a DRAFT . I still need to add the decorator to most websites & perhaps, ,idk, adjust tests?


27136773:
I don't think it is a good idea bringing in another interface function which needs a per-website basis kind of maintenance.

If at all a manga website search would be in higher demand than to copy a chapter URL.

MikeZeDev:
Well, the thing is not really about "copying the url". Its also about "opening the chapter in internal browser" and by doing so it can help troubleshooting. "Copying the url" is just a byproduct of this.

About maintenance: Seriously speaking, the impressive majority of websites will use default decorator and wont require maintenance at all.

Wdym by "manga website search" btw?

web/src/engine/websites/AnimeSama.ts
```diff
@@ -36,4 +36,10 @@ export default class extends DecoratableMangaScraper {
         const data = await FetchWindowScript<string[]>(new Request(new URL(`${chapter.Parent.Identifier}scan/vf/`, this.URI)), pageScript(parseInt(chapter.Identifier)), 500);
```
27136773, line 36: Avoid duplicate code, use `this.GetChapterURL` instead
(other places as well)

web/src/engine/websites/CMangax.ts
```diff
@@ -34,7 +35,21 @@ type APIPages = {
...
+    id: string;
+    slug: string;
+};
+
+type ChapterID = {
+    id: string;
+    num: string;
+};
+
+@Common.MangaCSS(/^https:\/\/cmangax\d+.com\/album\/[^/]+-\d+$/, 'h1 p.name', (paragraph, uri) => {
```
27136773: This seems to increase complexity just to be able getting the chapter URL.
Can't the URL be derived without overloading the manga/chapter ID?

MikeZeDev: Yes but that would mean a request to gather manga slug

MikeZeDev: Done

web/src/engine/websites/AnimeSama.ts
MikeZeDev: nice catch

27136773:
> Well, the thing is not really about "copying the url". Its also about "opening the chapter in internal browser" and by doing so it can help troubleshooting. "Copying the url" is just a byproduct of this.

I don't think it will be helpful for troubleshooting since the URL used by HakuNeko is likely to be different from the URL to show the chapters in the browser.

> About maintenance: Seriously speaking, the impressive majority of websites will use default decorator and wont require maintenance at all.

The current trend is that supported websites are updated to use Rest API endpoints instead of browser URLs. If this trend continues the additional maintenance burden is real.

> Wdym by "manga website search" btw?

Users always ask why their search is not working, because they think it is a live search on the website and not in the synchronized manga list. I barley see any user asking to open a specific chapter in the internal browser. So I still think a feature to get chapter URLs has a lower priority compared to other features with similar implementation costs (such as a live manga search on website).

PR #1722, opened 2026-06-15
engine: fix npm run test:websites

MikeZeDev:


web/src/frontend/classic/FrontendClassic_fixture.ts
```diff
@@ -30,7 +30,7 @@ export class TestFixture extends FrontendFixture {
      * Click the option in the website selection list that corresponds to the given website plugin {@link id}.
      */
     public async SelectWebsite(id: string, timeout = 1000) {
-        const selector = `#Plugin [role="listbox"] #${id}[role="option"] .title`;
+        const selector = `#Plugin #menu-PluginSelect #PluginSelect-${id}[role="option"] .title`;
```
27136773: @copilot How derive these CSS selectors?

web/test/FrontendFixture.ts
```diff
@@ -7,6 +7,11 @@ import type { IValue } from '../src/engine/SettingsManager';
  */
 export class FrontendFixture extends PuppeteerFixture {
 
+    static readonly #frontendSelectors: Readonly<Record<string, string>> = {
+        classic: '#PluginSelect',
```
27136773: @copilot Why using a random element instead of root element for svelte app?

web/test/FrontendFixture.ts
```diff
@@ -37,7 +43,8 @@ export class FrontendFixture extends PuppeteerFixture {
             page.off('dialog', dismiss);
         }
         await page.reload();
-        await this.Delay(500);
+        await page.waitForFunction(() => document.readyState === 'complete');
+        await page.waitForSelector(selector, { timeout: 15000, visible: true });
```
27136773: @copilot Why adding a wait when all wrappers are already have a waiting mechanism?

27136773:
@copilot Why making changes which are not related to the cause of this PR?

package.json
```diff
@@ -15,6 +15,7 @@
         "@stylistic/eslint-plugin": "5.10.0",
         "@types/chrome": "0.1.43",
         "@types/nw.js": "0.92.0",
+        "esbuild": "^0.28.1",
```
27136773, line 18: @copilot Why does this solve issue with vite on windows, but is not required on linux and macos

27136773: @copilot Propose a plan to migrate vite from rollup to rolldown instead of manually pulling in esbuild for legacy support

27136773: @copilot Implement the proposed plan with a fresh PR from main branch and add me as reviewer

27136773:
By migrating to Rolldown in e42f24ab the peer-dependency to esbuild should no longer be required

Issue #1778, opened 2026-08-02
[Bug]  "A requested file or directory could not be found " when importing bookmarks or downloading chapters


YvanRobR:
Both bookmark import and chapter download fail with a NotFoundError DOM exception. This affects all disk write/read operations, not just bookmarks — chapter downloads fail with the same error, so both symptoms point to the same root cause.

Console output when reproducing:

Unsafe attempt to load URL https://todo.com/ from frame with URL chrome-error://chromewebdata/. Domains, protocols and ports must match.

The literal todo.com placeholder-looking URL suggests a possible leftover/placeholder value in the build that never got replaced with a real URL, though I can't confirm this is the actual cause.

To Reproduce

Install hakuneko-electron via snap, edge channel (revision 8, 42.5.0)
Launch the app
Try to import a .bookmarks file, or download any chapter
Observe error toast: "NotFoundError: A requested file or directory could not be found at the time an operation was processed."

Elimination tests performed

To rule out sandbox/environment causes, I tested:

Running with sudo snap install hakuneko-electron --edge --devmode (confinement fully disabled) → same error, still occurs
Launching with GTK_USE_PORTAL=0 snap run hakuneko-electron → same error
Launching with snap run hakuneko-electron --ozone-platform=x11 → same error
Combination of the above two → same error
Checked journalctl -f | grep -i apparmor while reproducing → no AppArmor denials logged at the time of the error
Verified snap home interface is connected (snap connections hakuneko-electron)
Verified target directories (~/Downloads) exist with correct ownership/permissions

None of these changed the outcome, which points to an application-level bug rather than a snap confinement / Wayland / xdg-desktop-portal issue.

Comparison

Installed the Flatpak build (io.github.hakuneko.HakuNeko, an older version) on the same machine — bookmark import and downloads work fine there, but that build is old enough that several sites/connectors I use no longer work with it.

Context

HakuNeko version: snap edge, 42.5.0, revision 8 (2026-06-29)
OS: Ubuntu 26.04, fresh install
Desktop: GNOME (Wayland)
Only revision available on the edge channel via snap info hakuneko-electron; stable/candidate/beta channels are empty, so I couldn't test a previous revision.

YvanRobR:
Thanks for moving it, sorry.

MikeZeDev:
What do you try to import?  show the bookmark file

What do you try to download? Please name Manga chapter website ??

Error is perhaps similar but  i guess its unrelated.

[...]

PR #1788, opened 2026-08-08
fix: add desktop entry to snap package (media directory picker broken)

### LINKED ISSUE #1359: Cannot set media directory on Snap image [Linux]
I cant set manga directory, ive tried click the "media directory" but it show nothing.

Image

vitorwilson:
## Problem

On the snap package, clicking **Media directory** in settings does nothing (`showDirectoryPicker()` instantly rejects with `AbortError`). Reported in #1359.

Root cause chain (verified on the installed snap, rev 8):

1. The snap ships **no desktop entry** — the `desktop:` key in the generated `snapcraft.yaml` was a commented-out `# TODO: Create desktop entry`, so `meta/gui/` ends up empty.
2. `snap routine portal-info <pid>` therefore omits the `DesktopFile` key:

   ```
   [Snap Info]
   InstanceName=hakuneko-electron
   HasNetworkStatus=false
   ```

3. xdg-desktop-portal requires that key to register the snap and **denies all portal requests** from the app. Reproduced from inside the sandbox:

   ```
   $ snap run --shell hakuneko-electron -c 'gdbus call --session \
       --dest org.freedesktop.portal.Desktop \
       --object-path /org/freedesktop/portal/desktop \
       --method org.freedesktop.portal.FileChooser.OpenFile "" "t" "{}"'
   Error: GDBus.Error:org.freedesktop.DBus.Error.AccessDenied: Portal operation not allowed:
   Key file does not have key “DesktopFile” in group “Snap Info”
   ```

   (same error visible in the app log as `Gdk-WARNING: Failed to read portal settings`)

4. Chromium's `file_system_chooser` gets the failed method call → cancels the chooser → `AbortError`, so the click silently does nothing.

## Fix

Generate `snap/gui/<name>.desktop` during bundling and enable the `desktop:` key in `snapcraft.yaml`, so the snap gets `meta/gui/hakuneko-electron.desktop` and the portal can register it.

Verified end-to-end on the installed snap by simulating exactly this (a shim that injects `DesktopFile` into the `portal-info` output): after that, the same `FileChooser.OpenFile` call succeeds and `showDirectoryPicker()` opens the dialog and resolves with the picked directory.

## Notes

- Fixes #1359
- Not built locally (bundling requires `sudo snapcraft pack` + uploads to edge); the change is confined to the bundling script and only adds the desktop entry the TODO already anticipated.


Sheepux:
Did you actually test it working before/after ?
No blind AI coding please

vitorwilson:
> Did you actually test it working before/after ?
> No blind AI coding please

I did, I had to make this change because the application wasn't working on my machine. I could not select an output folder.

Sheepux:
We'll wait for @27136773 to review as he worked on the snap

MikeZeDev:
May also fix https://github.com/manga-download/haruneko/issues/1778 ?

app/electron/scripts/bundle-app-snap.mjs
```diff
@@ -35,15 +35,37 @@ async function createSnapImage(blinkDeploymentTemporaryDirectory, blinkDeploymen
...
     }
 }
 
+async function createDesktopEntry(blinkDeploymentOutputDirectory) {
+    const directory = path.join(blinkDeploymentOutputDirectory, 'snap', 'gui');
+    await fs.mkdir(directory, { recursive: true });
+    const file = path.join(directory, `${pkgConfig.name}.desktop`);
+    // A desktop entry is mandatory for xdg-desktop-portal to register the snap,
+    // otherwise all portal requests (e.g. the file chooser) are denied.
+    await fs.writeFile(file, [
```
27136773: Align with https://github.com/manga-download/hakuneko/blob/master/build-app.js#L400-L41 and https://github.com/manga-download/hakuneko/blob/master/build-app.config

27136773: Wrap in backticks instead of string `join`

app/electron/scripts/bundle-app-snap.mjs
```diff
@@ -52,15 +52,17 @@ async function createDesktopEntry(blinkDeploymentOutputDirectory) {
...
-Exec=${pkgConfig.name}
-Terminal=false
-Categories=Network;
-StartupNotify=true
-`);
+    await fs.writeFile(file, [
+        '[Desktop Entry]',
+        'Type=Application',
+        `Name=${pkgConfig.title}`,
+        `Comment=${pkgConfig.description}`,
```
27136773: Is this really necessary?

app/electron/scripts/bundle-app-snap.mjs
```diff
@@ -52,15 +52,17 @@ async function createDesktopEntry(blinkDeploymentOutputDirectory) {
...
+        `Exec=${pkgConfig.name}`,
+        'Terminal=false',
```
27136773: Is this really necessary?

app/electron/scripts/bundle-app-snap.mjs
```diff
@@ -52,15 +52,17 @@ async function createDesktopEntry(blinkDeploymentOutputDirectory) {
...
+        'Categories=Network;',
+        'StartupNotify=true',
```
27136773: Is this really necessary?

app/electron/scripts/bundle-app-snap.mjs
```diff
@@ -35,15 +35,35 @@ async function createSnapImage(blinkDeploymentTemporaryDirectory, blinkDeploymen
...
     const yaml = await createSnapcraftYaml(blinkDeploymentTemporaryDirectory, blinkDeploymentOutputDirectory);
+    const desktop = await createDesktopEntry(blinkDeploymentOutputDirectory);
     try {
         await run('sudo snapcraft pack --destructive-mode', blinkDeploymentOutputDirectory);
         await run(`sudo mv ${pkgConfig.name}*.snap ${snapfile}`, blinkDeploymentOutputDirectory);
         await run('snapcraft upload *.snap --release=edge', blinkDeploymentOutputDirectory);
     } finally {
         fs.unlink(yaml);
+        fs.rm(path.dirname(path.dirname(desktop)), { recursive: true, force: true });
     }
```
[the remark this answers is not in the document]
27136773: Cleanup mechanism should be addressed, but not as suggested by copilot

app/electron/scripts/bundle-app-snap.mjs
27136773: This is supposed to be a file not a directory, use appropriate options for deletion (`unlink`)

vitorwilson:
Review points addressed in 179d25af:

- **Align with hakuneko template**: the desktop entry now mirrors `hakuneko/build-app.js` + `build-app.config` — `Version=1.0`, `Type=Application`, `Name`, `GenericName`, `Exec`, `Icon`, `Categories=Network;FileTransfer;`.
- **Backticks instead of string join**: reverted to a template literal (this undoes a4b7c0b4's `.join()` conversion per the review request; note the join version also produced no trailing newline issues, but backticks match the requested style and the rest of this file).
- **`Comment` / `Terminal` / `StartupNotify` dropped** : not present in the reference template.
- **Cleanup**: removed the `try/finally` entirely; stale artifacts are now deleted up-front with the same awaited `fs.unlink` pattern already used for the `.snap` artifact (file-level `unlink`, no recursive `rm` on the directory). Since `createSnapcraftYaml`/`createDesktopEntry` overwrite their outputs, stale-file removal before generation is sufficient.
- Copilot's leading-whitespace concern: template literal content lines start at column 0, so generated output has no leading spaces (visible in the diff).

Generated `hakuneko-electron.desktop`:

```ini
[Desktop Entry]
Version=1.0
Type=Application
Name=HakuNeko
GenericName=Manga, Anime & Novel Downloader (Electron based desktop client)
Exec=hakuneko-electron
Icon=hakuneko-electron
Categories=Network;FileTransfer;
```

app/electron/scripts/bundle-app-snap.mjs
```diff
@@ -30,18 +30,36 @@ async function updateBinary(blinkApplicationResourcesDirectory, blinkDeploymentT
 
 async function createSnapImage(blinkDeploymentTemporaryDirectory, blinkDeploymentOutputDirectory) {
     const snapfile = path.basename(blinkDeploymentTemporaryDirectory).replace(/^electron/i, pkgConfig.name) + '.snap';
+    const yaml = path.join(blinkDeploymentOutputDirectory, 'snapcraft.yaml');
+    const desktop = path.join(blinkDeploymentOutputDirectory, 'snap', 'gui', `${pkgConfig.name}.desktop`);
     try {
-        const artifact = path.join(blinkDeploymentOutputDirectory, snapfile);
-        await fs.unlink(artifact);
+        await fs.unlink(path.join(blinkDeploymentOutputDirectory, snapfile));
```
27136773: When this `unlink` fails and throws then the following files are not correctly unlinked

app/electron/scripts/bundle-app-snap.mjs
```diff
@@ -30,18 +30,36 @@ async function updateBinary(blinkApplicationResourcesDirectory, blinkDeploymentT
...
+    // A desktop entry is mandatory for xdg-desktop-portal to register the snap,
+    // otherwise all portal requests (e.g. the file chooser) are denied.
+    // Field semantics follow hakuneko/build-app.config (meta.type, meta.categories).
+    await fs.writeFile(file, `[Desktop Entry]
+Version=1.0
+Type=Application
+Name=${pkgConfig.title}
+GenericName=${pkgConfig.description}
+Exec=${pkgConfig.name}
+Icon=${pkgConfig.name}
```
27136773: Shouldn't the icon be a path to a file?

vitorwilson:
Both points are addressed in `a1450fba`:

- Each artifact now has its own try/catch, so one failing unlink no longer skips the remaining deletions. The new icon file is covered as well.
- The icon is now shipped in `snap/gui/` (copied from `app/res/linux/icon.png`) and referenced by its installed absolute path, `Icon=${SNAP}/meta/gui/hakuneko-electron.png`, following the snapcraft documentation: https://ubuntu.com/docs/snapcraft/latest/how-to/crafting/configure-package-information/

app/electron/scripts/bundle-app-snap.mjs
```diff
@@ -28,20 +28,41 @@ async function updateBinary(blinkApplicationResourcesDirectory, blinkDeploymentT
...
-        await run('sudo snapcraft pack --destructive-mode', blinkDeploymentOutputDirectory);
-        await run(`sudo mv ${pkgConfig.name}*.snap ${snapfile}`, blinkDeploymentOutputDirectory);
-        await run('snapcraft upload *.snap --release=edge', blinkDeploymentOutputDirectory);
-    } finally {
-        fs.unlink(yaml);
-    }
+    const yaml = path.join(blinkDeploymentOutputDirectory, 'snapcraft.yaml');
+    const desktop = path.join(blinkDeploymentOutputDirectory, 'snap', 'gui', `${pkgConfig.name}.desktop`);
+    const icon = path.join(blinkDeploymentOutputDirectory, 'snap', 'gui', `${pkgConfig.name}.png`);
+    try { await fs.unlink(path.join(blinkDeploymentOutputDirectory, snapfile)); } catch { }
```
27136773: DRY: Consider using promises instead of repetitive `try .. .catch` blocks such as:
```ts
await Promise.allSettled([yaml, desktop, icon, ...].map(file => fs.unlink(file)));
```

vitorwilson: Thanks, I did not consider that.

vitorwilson: Applied in `cb272310`: all four unlinks now go through `Promise.allSettled([...].map(file => fs.unlink(file)))`, with the `path.join` kept inline for `snapfile` (still a basename for the `sudo mv` below).

27136773:
From a brief view this looks good.
Unfortunately I didn't had the chance to verify a Snap build yet before merging ...

MikeZeDev:
'Failed to generate desktop file "snap/gui/hakuneko-electron.desktop": file does no exist (defined in app "hakuneko-electron" when i try to build the snap using this. Ubuntu  24.04 LTS.

Removing the "desktop" entry from the generated yaml fixes the build process. Snapcraft is taking care of destktop and picture file by itself.

MikeZeDev:
I can confirm adding a .desktop entry restored the abillity to chose a manga folder ! Same OS than OP, with the same bug.

image


Now, perhaps it needs to be tweaked to work in Github CI

MikeZeDev:
@vitorwilson  Can you update from master so we can run validations pls?

PR #1795, opened 2026-08-15
Add LeerManhwas website connector

JuanManuelCastelo:
Adds support for https://leermanhwas.com, a spanish manhwa website.

The site runs on a custom platform instead of a known CMS, so the connector is composed from the existing common decorators. Its chapter anchors embed an update badge (a `NEW` image or the release date) next to the chapter name, which is dropped through an `AnchorInfoExtractor` to keep the chapter titles clean.

The manga list is scraped from the paginated `/page/{page}/` view, which is
the only paginated listing the site exposes. Happy to switch if there's a
better catalogue route.

Verified with `npm run test:websites -- LeerManhwas`: 5/5 passing against the
live site, including the image blob download.


web/src/engine/websites/LeerManhwas.ts
```diff
@@ -0,0 +1,24 @@
+import { Tags } from '../Tags';
+import icon from './LeerManhwas.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as Common from './decorators/Common';
+
+// NOTE: The chapter anchor contains an update badge (either a `NEW` image or the release date)
+//       next to the chapter name, which must be dropped to get a clean chapter title.
+const chapterInfoExtractor = Common.AnchorInfoExtractor(false, 'span.ct-update');
```
MikeZeDev: No need for a variable, use in decorator directly

web/src/engine/websites/LeerManhwas.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@Common.MangaCSS(/^{origin}\/manhwa\/[^/]+\/$/, 'h1.main-info-title')
```
MikeZeDev, line 6: Since that madara like website tends to put badges in titles, prefer using the "breadcrumb" on top.

ol.bread-crumb li:last-of-type

web/src/engine/websites/LeerManhwas_e2e.ts
```diff
@@ -0,0 +1,22 @@
...
+        id: '/manhwa/la-tumba-del-cisne/',
+        title: 'La tumba del cisne'
+    },
+    child: {
+        id: '/manhwa/la-tumba-del-cisne/capitulo-1/',
+        title: 'Capítulo 1'
+    },
+    entry: {
+        index: 0,
+        size: 504_665,
```
MikeZeDev: Not a big deal: The chosen picture is not really representative of the chapter since its used in others. May want to avoid that and if possible, also dont use aggregators pics. Well, dont worry, no need /not possible here :).

Suggestion : 

```ts
    child: {
        id: '/manhwa/la-tumba-del-cisne/capitulo-51/',
        title: 'Capítulo 51'
    },
    entry: {
        index: 5,
        size: 460_986,
    }

```

If possible try to minimize the bandwidth use. Your previous choice were fine, i had to chose image 5 because the others are bigger. Really not a big deal, just some optimal values.

MikeZeDev (changes requested):

MikeZeDev:
Thanks <3

PR #1800, opened 2026-08-17
Add NextJS helpers

MikeZeDev:
* Improved extraction of _next_f payloads, particularly nested ones
* Add helper for static extraction of `__NEXT_DATA__ `props.


web/src/engine/platform/FetchProviderCommon.ts
```diff
@@ -5,6 +5,7 @@ import { CreateRemoteBrowserWindow } from './RemoteBrowserWindow';
 import { CheckAntiScrapingDetection, FetchRedirection } from './AntiScrapingDetection';
 import type { FeatureFlags } from '../FeatureFlags';
 import { Delay, SetTimeout, ClearTimeout } from '../BackgroundTimers';
+import { ExtractNextJS } from '../NextJS';
```
27136773: Why outsourcing this, I don't see any benefit?
If this is a design decision, consider outsourcing all other Fetch methods as well into dedicated files.

web/src/engine/NextJS.ts
```diff
@@ -0,0 +1,310 @@
+// Regular expression to match Next.js App Router flight push chunks
+const NEXT_F_REGEX = /self\.__next_f\.push\(\s*(\[.*])\s*\)\s*;?\s*$/s;
+
+/**
+ * Extracts a matched value from a Next.js HTML document using a predicate.
+ */
+export function ExtractNextJS<T extends JSONElement>(doc: Document, predicate: (data: JSONObject<JSONElement> | JSONArray<JSONElement>) => unknown): T | undefined {
```
27136773: Wouldn't it be more beneficial to investigate how NextJS works under the hood and if the (de-)hydration library from NextJS is available and could be added as dependency instead of re-implementing the NExtJS (de-)hydration?

web/src/engine/platform/FetchProviderCommon.ts
MikeZeDev: Yes its by design. Because its quite complex, so much that i this it deserve a specific file. Because its not much "fetching" than "extracting", so semantically its not good. 
To me, other extraction methods still have their place in FetchProviderCommon.

Like you, i have no love for big units, hence we dont put everything in Common :D

Also, perphaps we could include helpers for the old __NEXT__.props next instead of having them in separate websites code. Maybe include and RSC extractor if needed in the future?

web/src/engine/NextJS.ts
MikeZeDev: NextJS doesnt expose methods to properly extract the data.
People have done open source parsers indeed, but in python :/

Well, maybe ill check them

MikeZeDev: * The python code i investigated wasnt able to parse nested payload.
* but THIS https://github.com/alcoceba/next.js-explorer seems kinda accurate :)

27136773:
Can `import { createFromReadableStream } from 'react-server-dom-webpack/client.browser';` package be used to create a React Component Tree directly from the Flight Data?

MikeZeDev:
> Can `import { createFromReadableStream } from 'react-server-dom-webpack/client.browser';` package be used to create a React Component Tree directly from the Flight Data?

Nope. 

* createFromReadableStream input must be "RSC binary stream" why is already transformed into "pseudo json" in the html/js of the page.

 * use() — Only works inside a React component during render (requires React context)

PR #1813, opened 2026-08-31
fix: YuriGarden - fix pages

anenasa:
JSON.parse is also used for other data that contains pages key. Only dispatch event if pages is the only key.


MikeZeDev:
Thanks for contribution.

PR #1816, opened 2026-09-02
fix: EighteenComic - reverse chapters order

anenasa:


MikeZeDev:
I reverted this PR because in this particular case, we are calling `FetchChaptersSinglePageCSS `before.
In my incoming PR i extended `FetchChaptersSinglePageCSS `with an optional parameter to reverse output order.

Since there is a need for that in more than a couple of website, changing `FetchChaptersSinglePageCSS ` and `ChaptersSinglePageCSS` are the way to go ;)

PR #1829, opened 2026-09-11
fix: Fix stalled fetches causing download list to hang forever

### LINKED ISSUE #1827: Download queue is stuck/stalled and unable to queue new download
# Description
When fetching items in download queue, request exception can lead to next download item in the queue to be stuck forever and unable to queue new download item.

# Screenshot 
As you can see, we're stuck at item 24 and I am unable to queue new download items, even though item 24 appears as completed. The queue won't process or accept new download items at all. The only way for the app to work again is to restart it, which lose all in memory download queue.

Image

# Pull request
- There's a fix here in my forked repo that was done with Codex, and I could verify the download item is no longer stuck after the fix. The error ones can be retried as normal, and the queue can continue processing download items. 

I can create a PR if we think this can be fixed. Verified with my local build.

https://github.com/manga-download/haruneko/compare/master...hieuxlu:haruneko:fix/stalled-fetches?expand=1

hieuxlu:
Closes https://github.com/manga-download/haruneko/issues/1827

### DIFF
--- a/web/src/engine/platform/FetchProviderCommon.ts
+++ b/web/src/engine/platform/FetchProviderCommon.ts
@@ -271,10 +271,20 @@ export abstract class FetchProvider {
         };
 
         return new Promise<T>((resolve, reject) => {
-            let cancellation = SetTimeout(async () => {
+            let settled = false;
+            let cancellation: Promise<number>;
+
+            const settle = async (action: () => void) => {
+                if (settled) return;
+                settled = true;
+                ClearTimeout(await cancellation);
                 await destroy();
+                action();
+            };
+
+            cancellation = SetTimeout(() => settle(() => {
                 reject(new Exception(R.FetchProvider_FetchWindow_TimeoutError));
-            }, timeout);
+            }), timeout);
 
             win.DOMReady.Subscribe(async () => {
                 invocations.push({ name: 'DOMReady', info: `Window: ${win}` });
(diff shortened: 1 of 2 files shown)


web/src/engine/platform/FetchProviderCommon_test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+function CreateWindow(options: { openError?: Error, scriptError?: Error } = {}) {
+    const domReady = new Observable<void, IRemoteBrowserWindow>(null);
+    const window = {
+        DOMReady: domReady,
+        BeforeWindowNavigate: new Observable<URL, IRemoteBrowserWindow>(null),
+        BeforeFrameNavigate: new Observable<URL, IRemoteBrowserWindow>(null),
+        Open: vi.fn(async () => {
+            if (options.openError) throw options.openError;
+            queueMicrotask(() => domReady.Dispatch());
```
27136773, line 35: Function `queueMicrotask` seems not to be declared anywhere

web/src/engine/platform/FetchProviderCommon_test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+class TestFetchProvider extends FetchProvider {
+    public override Fetch(request: Request): Promise<Response> {
+        return fetch(request);
+    }
+}
+
+function CreateWindow(options: { openError?: Error, scriptError?: Error } = {}) {
+    const domReady = new Observable<void, IRemoteBrowserWindow>(null);
+    const window = {
```
27136773, line 29: This mock is way to complex and therefore error prone.
Consider writing end-to-end test with real browser window instead.

web/src/engine/platform/FetchProviderCommon.ts
```diff
@@ -296,21 +306,21 @@ export abstract class FetchProvider {
...
-                } catch {
-                    await destroy();
+                } catch (error) {
+                    await settle(() => reject(error));
                 }
             });
 
             invocations.push({ name: 'Open', info: `Request URL: ${request.url}` });
-            win.Open(request, this.featureFlags.VerboseFetchWindow.Value, preload);
+            win.Open(request, this.featureFlags.VerboseFetchWindow.Value, preload).catch(error => {
```
27136773, line 321: Is there a use case when the `Open` failed?

web/src/engine/platform/FetchProviderCommon.ts
```diff
@@ -271,10 +271,20 @@ export abstract class FetchProvider {
         };
 
         return new Promise<T>((resolve, reject) => {
-            let cancellation = SetTimeout(async () => {
+            let settled = false;
+            let cancellation: Promise<number>;
+
+            const settle = async (action: () => void) => {
+                if (settled) return;
```
27136773, line 278: Is this guarding really necessary?
Function `ClearTimeout` is robust and should handle multiple calls and errors just fine.

PR #1830, opened 2026-09-11
fix VIZ reader authentication and stalled fetches

### LINKED ISSUE #1828: [viz.com plugin] Too many request attempts (due to using legacy single page API)
# Description
viz.com plugin regularly responds with 429 (Too Many Requests) status code when downloading chapters. 

After working with Codex, I found this is due to the plugin using legacy single page API that's no longer used by the website any more. The viz.com website is using multiple pages API that would greatly improve the performance and reduce rate limiting chances.

# Screenshot
Example of 429 response after downloading ~10 chapters and lots of single page API request
Image

# Pull request
This proposes to use what viz.com is website currently using by fetching all page URLs using a single call in page script, and return the CDN pageUrls to the client to avoid rate limiting. Verified working with local build with no rate limiting. 

Please let me know if this can be merged/fixed. I understand that only users with viz.com subscription can verify this, which I do have.

There's a separate but necessary changes that include all URL partitioned cookies to support authentication of sub pages
https://github.com/manga-download/haruneko/compare/master...hieuxlu:fix/viz-reader-authentication?expand=1

# After fix

Screenshot after fix, a single API call that retrieve all pageUrls are done in main Pagescript and the client only need to pull the pageUrls from CDN directly, avoid rate limiting.

Image

hieuxlu:
Closes https://github.com/manga-download/haruneko/issues/1828

### DIFF
--- a/web/src/engine/platform/electron/FetchProvider_test.ts
+++ b/web/src/engine/platform/electron/FetchProvider_test.ts
@@ -72,6 +72,25 @@ describe('FetchProvider', () => {
 
     describe('Fetch', () => {
 
+        it('Should request cookies for the complete request URL', async () => {
+            const fixture = new TestFixture();
+            const testee = fixture.CreateTestee();
+            testee.Initialize({} as FeatureFlags);
+            const request = new Request('https://www.viz.com/manga/get_manga_url?page=1');
+            const fetch = vi.spyOn(globalThis, 'fetch').mockResolvedValue(new Response());
+
+            try {
+                await testee.Fetch(request);
+            } finally {
+                fetch.mockRestore();
+            }
+
+            expect(fixture.MockIpcRenderer.invoke).toHaveBeenCalledWith(
+                Channels.FetchProvider.GetSessionCookies,
+                { url: request.url }
+            );
+        });
+
         it('Should passthru GET to native fetch', async () => {
             const fixture = new TestFixture();
             const testee = fixture.CreateTestee();
--- a/web/src/engine/websites/VizShonenJump.ts
+++ b/web/src/engine/websites/VizShonenJump.ts
@@ -15,6 +15,8 @@ import { WebsiteResourceKey as R } from '../../i18n/ILocale';
 type PagesInfos = {
     pagesCount: number,
     mangaID: string;
+    pageURLs: string[];
+    error?: string;
 };
 
 type ExifData = {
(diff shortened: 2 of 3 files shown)


MikeZeDev:
First : thanks for pointing out that we can chain image id in one request. This has been taken care of in https://github.com/manga-download/haruneko/commits/master/. 👌 

Please update your branch according to it.

Second : we arent gonna modify Fetch* units without a solid understanding off the problem / if not needed.
The 403 errors you got were most likely something related to requests "headers", not auth.

Let's be pragmatic.

Try with our last fix, and only our last fix. Not your PR. If you still has real auth problems, we will discuss further 👌

web/src/engine/platform/electron/FetchProvider_test.ts
```diff
@@ -72,6 +72,25 @@ describe('FetchProvider', () => {
 
     describe('Fetch', () => {
 
+        it('Should request cookies for the complete request URL', async () => {
```
MikeZeDev, line 75: Bad idea in any case.. Viz is Geoblocked, test may fail

web/src/engine/websites/VizShonenJump.ts
```diff
@@ -90,26 +124,12 @@ export default class extends DecoratableMangaScraper {
...
-    private async TestAccessAndDummyPage(pages: Page[]): Promise<Page[]> {
-        // If chapter is not accessible : 'url' wont be an url at all => throw Plugin_Common_Chapter_UnavailableError.
-        const lastPage = pages.at(-1);
-        const url = await (await Fetch(new Request(lastPage.Link, { headers: { Referer: lastPage.Parameters.Referer, } }))).text();
-        if (!url.startsWith('http')) throw new Exception(R.Plugin_Common_Chapter_UnavailableError);
-
-        // last page may be a dummy (unavailable) page. In that case strip it from page array.
-        const response = await fetch(new Request(url, { method: 'HEAD', headers: { Referer: lastPage.Parameters.Referer } }));
-        return response.status != 403 ? pages: pages.slice(0, -1);
+        const { pageURLs, error } = await FetchWindowScript<PagesInfos>(new Request(chapterurl), PagesScript, 1500);
```
MikeZeDev, line 127: you removed the "last page test". It exists for a reason !

See the test units. Last page of  Jojo part 5 chapter 3 always return 403 in a REAL BROWSER. But Viz reader doesnt care and if you check elswhere on the net this page doesnt exist.

hieuxlu:
Thanks, looks like the bulk page load is committed upstream so I'm good with closing this

PR #1831, opened 2026-09-12
feat(connectors): add Inkapk, Manga Portali, and Violet Scans

sagebeme:
Closes #8693, closes #8701, closes #8679.

## Inkapk (inkapk.net) -- closes #8693

Standard Madara theme (verified via the theme's `admin-ajax.php` `madara_load_more` listing endpoint), plain `WordPressMadara` decorator subclass, no overrides needed.

## Manga Portalı (mangaportali.com) -- closes #8701

Custom Next.js site, fully server-rendered, no API reverse-engineering needed. The manga page has "first chapter"/"latest chapter" shortcut buttons linking to the same URLs as two entries in the real chapter list -- `FetchChapters` dedupes those, keeping the real list's title text. Note: could only find a single server-rendered page of `/series` (20 series) -- no page-based pagination worked, so there may be more behind client-side infinite scroll or an API call not identified. Flagging that rather than claiming full coverage.

## Violet Scans (violetscans.org) -- closes #8679

Mangastream theme with a customized listing path (`/comics/` instead of the template default `/list/`) and pagination, so `FetchMangas` uses `Common.MangasMultiPageCSS` directly. Also has a locked/paid chapter system: locked chapters render as a Bootstrap modal trigger (`data-bs-toggle="modal"`) with no `href` at all rather than a real link, so `ChaptersSinglePageCSS`'s `a[href]` selector naturally excludes them -- not a special-cased paywall bypass.

## Testing

All three checked against live responses before submitting: the manga listing page, a full chapter list (Violet Scans included a mix of free and locked chapters to confirm the locked ones are correctly skipped), and the page-image URLs from one real chapter per site.

PR #1833, opened 2026-09-14
feat: add Inkapk, MangaPortali, VioletScans

### LINKED ISSUE #8679: [Site Request] Violet Scans Manhwa website
### Is there an existing request for this website?

- [x] I have searched the existing issues

### Website Name

Violet Scans

### Site URL

https://violetscans.org/

### Series list

https://violetscans.org/comics/

### Manga example

https://violetscans.org/comics/101st-confession/

### Manga chapter viewer

https://violetscans.org/101st-confession-chapter-69/

### Languages

English

### Website relationship

_No response_

### Additional details

Please add this website also...

### LINKED ISSUE #8693: [Site Request] Inkapk
### Is there an existing request for this website?

- [x] I have searched the existing issues

### Website Name

Inkapk

### Site URL

https://inkapk.net/

### Series list

https://inkapk.net/obras/

### Manga example

https://inkapk.net/obras/wireless-onahole/

### Manga chapter viewer

https://inkapk.net/obras/wireless-onahole/capitulo-98/

### Languages

Brazilian Portuguese

### Website relationship

_No response_

### Additional details

_No response_

### LINKED ISSUE #8701: [Site Request] mangaportali
### Is there an existing request for this website?

- [x] I have searched the existing issues

### Website Name

mangaportali

### Site URL

https://www.mangaportali.com

### Series list

https://www.mangaportali.com/series

### Manga example

https://www.mangaportali.com/series/solo-leveling

### Manga chapter viewer

https://www.mangaportali.com/reader/solo-leveling/bolum-200

### Languages

Turkey

### Website relationship

_No response_

### Additional details

_No response_

sagebeme:
Three sites, one commit each, each with an e2e test that downloads a real page and checks its size.

- Inkapk: closes https://github.com/manga-download/hakuneko/issues/8693
- MangaPortali: closes https://github.com/manga-download/hakuneko/issues/8701
- VioletScans: closes https://github.com/manga-download/hakuneko/issues/8679

Supersedes #1831 - same code, restructured into one commit per site with correct issue links.

Thank you for the review on the first attempt. The pointers to the site's actual API, the list-mode page, and writing a proper e2e test taught me a lot I didn't know going in.


web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        type This = typeof this;
+        return Array.fromAsync(async function* (this: This) {
+            for (let page = 1, totalPages = 1; page <= totalPages; page++) {
+                const uri = new URL('/api/series', this.URI);
```
MikeZeDev: * Prefer adding a readonly class variable "apiURL" and use it in requests. Make sure it includes trailing "/".
*  No need to multiply lines, and should be using deconstruction. We would rather do it like that

```ts
const { objectproperty } = await FetchJSON<type>(new Request(new URL(`./series/`,this.apiURL);
```

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+                uri.searchParams.set('page', `${page}`);
+                uri.searchParams.set('pageSize', '20');
```
MikeZeDev: * Parameters should be directly appended when doing "new URL(...") , using interpolation `` .
we use searchparams.set only when there are url encoding concerns

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+                const data = await FetchJSON<APIPaginated<APIEntry>>(new Request(uri));
```
MikeZeDev: Please use deconstruction to get the variables you need. So it consistent with the  map after.

```ts
const { totalPages, items }) = await FetchJSON....
```

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public override get Icon() {
+        return icon;
+    }
+
+    public override async FetchMangas(provider: MangaPlugin): Promise<Manga[]> {
+        type This = typeof this;
+        return Array.fromAsync(async function* (this: This) {
+            for (let page = 1, totalPages = 1; page <= totalPages; page++) {
```
MikeZeDev: Please respect the coding style. keep the run flag.
Just update the flag according to page and totalpages at the end of the loop.

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+                const uri = new URL('/api/series', this.URI);
+                uri.searchParams.set('page', `${page}`);
+                uri.searchParams.set('pageSize', '20');
+                const data = await FetchJSON<APIPaginated<APIEntry>>(new Request(uri));
+                totalPages = data.totalPages;
+                yield* data.items.map(({ slug, title }) => new Manga(this, provider, `/series/${slug}`, title));
```
MikeZeDev:
```ts
const mangas = items.map.....;
yield * mangas;
```

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }.call(this));
+    }
+
+    public override async FetchChapters(manga: Manga): Promise<Chapter[]> {
+        const slug = manga.Identifier.split('/').pop();
+        type This = typeof this;
+        return Array.fromAsync(async function* (this: This) {
```
MikeZeDev, line 53: basically, same remarks than above. All of them. Code refactoring :)

web/src/engine/websites/MangaPortali.ts
MikeZeDev: If there is no reason to retain the whole pathname of the manga just use slug as identifier.
And dont forget to adapt the use of Common.MangaCSS accordingly.

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+            for (let page = 1, totalPages = 1; page <= totalPages; page++) {
+                const uri = new URL(`/api/series/${slug}/chapters`, this.URI);
+                uri.searchParams.set('page', `${page}`);
+                const data = await FetchJSON<APIPaginated<APIEntry>>(new Request(uri));
+                totalPages = data.totalPages;
+                yield* data.items.map(chapter => new Chapter(this, manga, `/reader/${slug}/${chapter.slug}`, chapter.title));
```
MikeZeDev: If there is no reason to retain the whole pathname of the chapter just use slug as identifier.

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }.call(this)).then(chapters => chapters.reverse());
+    }
+
+    public override async FetchPages(chapter: Chapter): Promise<Page[]> {
+        const [ , , mangaSlug, chapterSlug ] = chapter.Identifier.split('/');
```
MikeZeDev: Providing you only used manga slug and chapter slug as respective identifiers like ive suggested, this can be greatly simplified.
Also please inline uri construction with FetchJSON call 

`        const { pages } = await FetchJSON<APIChapter>(new Request(new URI .....));`

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+        const uri = new URL(`/api/series/${mangaSlug}/chapters/${chapterSlug}`, this.URI);
+        const { pages } = await FetchJSON<APIChapter>(new Request(uri));
+        return pages
+            .sort((self, other) => self.index - other.index)
```
MikeZeDev: We are now enforcing Descending order sorting in engine. This is Ascending sorting .

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+            .map(page => new Page(this, chapter, new URL(page.imageUrl)));
```
MikeZeDev: `.map ({ imageUrl }) =>`

web/src/engine/websites/VioletScans.ts
```diff
@@ -0,0 +1,21 @@
+import { Tags } from '../Tags';
+import icon from './VioletScans.webp';
+import { DecoratableMangaScraper } from '../providers/MangaPlugin';
+import * as MangaStream from './decorators/WordPressMangaStream';
+import * as Common from './decorators/Common';
+
+@MangaStream.MangaCSS(/^{origin}\/comics\/[^/]+\/$/)
+@MangaStream.MangasSinglePageCSS(undefined, '/comics/list-mode/')
+@MangaStream.ChaptersSinglePageCSS('#chapterlist li a[href]')
```
MikeZeDev, line 9: They have "a" without href? Just curious.

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+        imageUrl: string;
+    }[];
+};
+
+@Common.MangaCSS(/^{origin}\/series\/[^/]+$/, 'h1')
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+
+    public constructor() {
+        super('mangaportali', 'Manga Portalı', 'https://www.mangaportali.com', Tags.Media.Manga, Tags.Language.Turkish, Tags.Source.Scanlator);
```
MikeZeDev: * Website says "Manga, Manhua, Manhwa". We miss some tags then.
* Only use `Scanlator` if you know they actually scanlate. Otherwise dont bother and default to "Aggregator".

MikeZeDev (dismissed):
This is way better 👌 .

This just need some minor code changes.

web/src/engine/websites/MangaPortali.ts
```diff
@@ -0,0 +1,72 @@
...
+        type This = typeof this;
+        return Array.fromAsync(async function* (this: This) {
+            for (let page = 1, totalPages = 1; page <= totalPages; page++) {
+                const uri = new URL(`/api/series/${slug}/chapters`, this.URI);
+                uri.searchParams.set('page', `${page}`);
+                const data = await FetchJSON<APIPaginated<APIEntry>>(new Request(uri));
+                totalPages = data.totalPages;
+                yield* data.items.map(chapter => new Chapter(this, manga, `/reader/${slug}/${chapter.slug}`, chapter.title));
+            }
+        }.call(this)).then(chapters => chapters.reverse());
```
MikeZeDev: * I assume API returns them in 01-> 999 order and thats why you reverse, right? 
 * Also dont use .then
```ts
        return ( await Array.fromAsync(async function* (this: This) {
.......
        }.call(this))).reverse();
```

web/src/engine/websites/Inkapk.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@Madara.MangaCSS(/^{origin}\/obras\/[^/]+\/$/, 'h1.ink-det-title')
+@Madara.MangasMultiPageAJAX()
+@Madara.ChaptersMultiPageAJAX()
+@Madara.PagesSinglePageCSS('div.reading-content img.wp-manga-chapter-img')
+@Common.ImageAjax()
+export default class extends DecoratableMangaScraper {
+
+    public constructor() {
+        super('inkapk', 'Inkapk', 'https://inkapk.net', Tags.Media.Manga, Tags.Language.Portuguese, Tags.Source.Scanlator);
```
MikeZeDev: They only got mangas? They are scanlators, not aggregators ?

web/src/engine/websites/VioletScans_e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+    },
+    container: {
+        url: 'https://violetscans.org/comics/101st-confession/',
+        id: '/comics/101st-confession/',
+        title: '101st Confession',
+    },
+    child: {
+        id: '/101st-confession-chapter-1/',
+        title: 'Chapter 1',
+        timeout: 20_000,
```
MikeZeDev: Website is so slow?

web/src/engine/websites/VioletScans_e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+    },
+    entry: {
+        index: 0,
+        size: 1_015_164,
+        type: 'image/webp',
+        timeout: 20_000,
```
MikeZeDev: timeout should not be necessary here. Its just fetching an image. Unless the image is like 500Mb or something.

web/src/engine/websites/MangaPortali.ts
MikeZeDev: * Try to increase pageSize if possible, so we will perform less requests.

MikeZeDev:
Please update from master so we can run test workflows :D

web/src/engine/websites/MangaPortali.ts
sagebeme: The API actually returns them newest first (Bölüm 200 → 101), so the reverse was wrong. Removed it, and dropped the .then too.

web/src/engine/websites/VioletScans_e2e.ts
sagebeme: Removed the image timeout, you're right.

web/src/engine/websites/VioletScans_e2e.ts
sagebeme: The site itself is fast (~1.3s per page), but the chapter step takes ~10-11s in the test harness and failed at the 7.5s default 3 out of 3 times, so I kept that one.

web/src/engine/websites/VioletScans.ts
sagebeme: Yes, those are coin-locked chapters. The <a> opens a purchase modal instead of linking, so a[href] skips them.

web/src/engine/websites/Inkapk.ts
sagebeme: Only manhwa, and they take work submissions through Discord, so I switched it to Manhwa + Aggregator.

web/src/engine/websites/MangaPortali.ts
sagebeme: Added Manga, Manhwa and Manhua and switched to Aggregator, nothing on the site says they translate.

web/src/engine/websites/MangaPortali.ts
sagebeme: Inlined the params. The API caps pageSize at 50 for series and 100 for chapters, so I'm using those.

sagebeme:
Thanks for the detailed review, I learned a lot from it. Everything is addressed in the last three commits: apiURL, run flag, destructuring, inline URLs, slugs as identifiers, and the tags. The e2e tests for all three sites pass locally (15/15).

sagebeme:
Updated from master.

web/src/engine/websites/VioletScans.ts
MikeZeDev: Our position is not fixed yet about locked chapters, i guess its fine 🆗

web/src/engine/websites/VioletScans_e2e.ts
MikeZeDev: we should be ok with 15_000. Oh well 🤷

web/src/engine/websites/MangaPortali.ts
MikeZeDev: Yeah we are enforcing desc sorting on the engine now.
* So the UI can work without problem with its own sorting.
* Because we assume that people want newest first
* Because.... its default for many website so its less work ;D

MikeZeDev:
Thanks for your contribution 👌

